PR #6, opened 2015-02-01
Make all existing tests run

TheLudd:
All existing tests can now be executed by running npm test. Mocha was
installed as a devDependency and is used by the npm test script.

The IO type depends on ramda and because of this ramda was installed
as a dependency. This could later be removed if the purpose is to make
this package dependency free.

A .gitignore file was created which includes the node_modules folder.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,1 @@
+node_modules/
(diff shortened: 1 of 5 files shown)


buzzdecafe:
LGTM

PR #7, opened 2015-02-01
Commonjsified all types

TheLudd:
All the existing types has been converted to the common js format. An
index file has been created that exposes makes them possible to import
like this:
```
var Maybe = require('ramda-fantasy').Maybe;
```

The tests now use these imports, all tests pass.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,0 +1,8 @@
+module.exports = {
+    Either: require('./src/Either'),
+    Future: require('./src/Future'),
+    Identity: require('./src/Identity'),
+    IO: require('./src/IO'),
+    Maybe: require('./src/Maybe'),
+    Reader: require('./src/Reader')
+};
--- a/src/Either.js
+++ b/src/Either.js
@@ -1,77 +1,61 @@
-(function(factory) {
-    if (typeof define === 'function' && define.amd) {
-        // AMD. Register as an anonymous module.
-        define(factory);
-    } else if (typeof exports === 'object') {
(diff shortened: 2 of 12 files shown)


davidchambers:
Should this pull request include a build step involving [Browserify][1] or [CommonJS Everywhere][2] so the code can be used in the browser?


[1]: https://github.com/substack/node-browserify
[2]: https://github.com/michaelficarra/commonjs-everywhere

CrossEye:
I believe we do want a [Browserify][br] step.

But first, we probably want to know what outputs we would want from our build script.  Is it something like

  * ramda.either.js
  * ramda.io.js
  * ramda.maybe.js
  * ramda.types.js (everything)

Or do we want some custom mix-and-match? 

  [br]: https://github.com/substack/node-browserify

TheLudd:
Is it not up to the user to determine how they want to browserify their dependencies? If we don't do it at all, we leave it entirely up to the end user to include all, one or some of the modules in their build.

TheLudd:
That being said, it is my personal preference to split up commits into as small pieces as possible. I see this commit as a small step forward no matter what build process we might want to use. And I think that this change, and creating the build process can preferably be split into two separate commits.

buzzdecafe:
> it is my personal preference to split up commits into as small pieces as possible.

I'm inclined to agree with this, so unless there are objections, let's pull this now and browserify later. Let's keep the ball rolling

PR #8, opened 2015-02-08
data Maybe a = Just a | Nothing

buzzdecafe:
updates to Maybe from recent discussion on fantasy-land here https://github.com/fantasyland/fantasy-land/issues/85

### DIFF
--- a/src/Maybe.js
+++ b/src/Maybe.js
@@ -1,45 +1,82 @@
-module.exports = Maybe;
+var util = require('./internal/util.js');
 
 function Maybe(x) {
-    if (!(this instanceof Maybe)) {
-        return new Maybe(x);
-    }
-    this.value = x;
+    throw new Error('Maybe cannot be directly instantiated.');
 };
 
-Maybe.of = function(x) {
-    return new Maybe(x);
+function _Just(x) {
+    this.value = x;
+}
+util.extend(_Just, Maybe);
+
+function _Nothing() {}
+util.extend(_Nothing, Maybe);
+
+var _nothing = new _Nothing();
+
+Maybe.Nothing = function() {
+    return _nothing;
 };
+
+Maybe.Just = function(x) {
+    return new _Just(x);
+};
+
+Maybe.fromNullable = function(x) {
+    return x == null ? _nothing : Maybe.Just(x);
+};
+
+Maybe.of = Maybe.Just;
+
+Maybe.prototype.of = Maybe.Just;
+
 
 // functor
-Maybe.prototype.map = function(f) {
-    return this.value == null ? this : new Maybe(f(this.value));
+_Just.prototype.map = function(f) {
+    return this.of(f(this.value));
 };
+
+_Nothing.prototype.map = util.returnThis;
 
 // apply
 // takes a Maybe that wraps a function (`app`) and applies its `map`
 // method to this Maybe's value, which must be a function.
-Maybe.prototype.ap = function(m) {
-    return typeof this.value !== 'function' ? new Maybe(null) : m.map(this.value);
+_Just.prototype.ap = function(m) {
+    return m.map(this.value);
 };
 
+_Nothing.prototype.ap = util.identity;
+
 // applicative
-Maybe.prototype.of = Maybe.of;
+// `of` inherited from `Maybe`
+
 
 // chain
 //  f must be a function which returns a value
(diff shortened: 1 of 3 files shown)


src/Maybe.js
```diff
@@ -1,45 +1,82 @@
-module.exports = Maybe;
+var util = require('./internal/util.js');
 
 function Maybe(x) {
-    if (!(this instanceof Maybe)) {
-        return new Maybe(x);
-    }
+    throw new Error('Maybe cannot be directly instantiated.');
```
TheLudd: I didn't understand why the main function couldn't return ```Just``` or ```Nothing```
```
function Maybe(x) {
    return x == null ? _nothing : Maybe.Just(x);
};
```

src/Maybe.js
```diff
@@ -1,45 +1,82 @@
...
 
 // monad
 // A value that implements the Monad specification must also implement the Applicative and Chain specifications.
 // see above.
 
 // equality method to enable testing
-Maybe.prototype.equals = function(that) {
-    return this.value === that.value;
+_Just.prototype.equals = function(that) {
+    return that.datatype = _Just && this.value === that.value;
```
davidchambers: You mean `===` here rather than `=`. Where's the linter? ;)

Why `that.datatype === _Just` rather than `that instanceof _Just`? Are you imagining a scenario in which maybes created by different instances of this module are used together?

src/Maybe.js
davidchambers: In Haskell terms `Maybe` is a type whereas `Just` and `Nothing` are data constructors. It doesn't make sense to create a `Maybe` directly as it's a set of data constructors which produce a value of the type: `Maybe = {Just, Nothing}`.

`Maybe.from` provides the old null semantics without making it impossible to represent `Just(null)` and `Just(undefined)`.

src/Maybe.js
buzzdecafe: > You mean `===` here rather than `=`.

oops. guess we need jshint.

> Why that.datatype === _Just rather than that instanceof _Just

i just don't like `instanceof` :smile: . but it would probably make more sense in case we ever want to "extend" these types

davidchambers:
This is a step in the right direction. :)

src/Maybe.js
TheLudd: But the ability to represent ```Just(null)``` that was apparently what ```Maybe.of``` was for.

buzzdecafe: `Just(null)` still feels pretty weird to me, but after the discussion over at fantasy-land, i am going in that direction.

TheLudd: @buzzdecafe Agreed, but my question here was about the constructor? Why throw an error instead of doing the null check? The constructor is not a part of the specification right?

TheLudd: @buzzdecafe You said you liked ```R.compose(R.map(f), Maybe, g);```.  I didn't see why we can't have it.

davidchambers: `R.compose(R.map(f), Maybe.from, g)` is more explicit.

src/Maybe.js
davidchambers: Your current approach supports "subclassing", since "subclasses" will inherit the `datatype` property.

src/Maybe.js
```diff
@@ -1,45 +1,82 @@
...
+    return _nothing;
 };
 
-Maybe.of = function(x) {
-    return new Maybe(x);
+Maybe.Just = function(x) {
+    return new _Just(x);
 };
 
+Maybe.from = function(x) {
```
TheLudd: Apparently ```from``` is a method on ```Comonad``` https://github.com/fantasyland/fantasy-land/issues/85#issuecomment-73523980

buzzdecafe: updating to `fromNullable`

src/Maybe.js
buzzdecafe: made it `fromNullable` because of collision with comonad spec

davidchambers:
:deciduous_tree:

CrossEye:
`fromNullable`?  Man, we need a better preposition!  `around`? `about`?  (And now I hear Yes, singing "Roundabout".  Life isn't fair, sometimes.)

buzzdecafe:
`thence`?

PR #9, opened 2015-02-09
test: simplify interface declaration

davidchambers:
This seems clearer to me.


CrossEye:
:fireworks: 

I think the point of the original was to demonstrate the inheritance tree, but that really got lost in the noise.  It's clearer in the refactored version.    Moreover it's even arguably clearer that `monad` inherits equally from `applicative` and `chain`, which was really lost in the original.  :+1:

buzzdecafe:
:cow2:

PR #12, opened 2015-02-16
revised Either, closes #11

### LINKED ISSUE #11: make `Either.fromNullable`; remove constructor impl
analogous to `Maybe`

buzzdecafe:


CrossEye:
I'm afraid I'm not seeing the point of these `fromNullable` conversions.

I saw the name mentioned as a possibility in one FL thread, but it didn't seem to become anything serious.  And it is seriously ugly compared to just `from`, which I know is taken now by Comonad, or using the type function with or without `new`.  Was there something fundamentally wrong with using the `Either` function directly to generate a Left or a Right, depending upon the inputs, or exposing functions to generate the Left and Right directly?

buzzdecafe:
There are exposed functions to generate `Left | Right` directly, viz `Either.Left | Either.Right`. This impl is in tune with the revised `Maybe`, which does not allow direct calling of `Maybe` constructor. I agree `fromNullable` is obnoxious, especially compared to `from`. But what else have we got?

CrossEye:
I guess I simply don't understand why what's here as `Either.fromNullable` can't be the implementation of `Either`.  It won't act as a true constructor, more as a factory, but I don't think that's an issue.

buzzdecafe:
IIUC the types should be able to support crazy sh*t like `Just(null)` and `Right(undefined)` in order to be fully lawful. Then the `fromNullable` biz is an interface that makes them behave how I can use ...

CrossEye:
Yes, I understand that.  That's what I got from that thread.  But that simply means that `of` has to support those.  And maybe this is the only way to go about doing so, but it shouldn't dictate that our standard usage using constructors or factories, which are not part of the spec, bothers with such esoterica.

buzzdecafe:
are you suggesting using the "constructor" instead of the `fromNullable`, more or less as I used to?

TheLudd:
@buzzdecafe I would support that. I even asked about that in the `Maybe` discussion but I never got a reply from you, @davidchambers said that `from` was more explicit and perhaps it is. But I think I prefer `Maybe(x)` and `Either(leftVal, x)` from `Maybe.fromNullable(x)` and `Either.fromNullable(leftVal, x)`. Especially since the `fromNullable` case would how I create my types in 99% of the cases.

src/Either.js
```diff
@@ -1,61 +1,69 @@
...
+util.extend(_Left, Either);
+
+_Left.prototype.ap = function(that) { return that; };
+
+Either.Left = function(value) {
+    return new _Left(value);
+};
+
+
+Either.fromNullable = function(left, right) {
```
TheLudd: Shouldn't this be curried? 
And if it is, do we decide to depend on `ramda` here? Perhaps it makes sense to have `ramda` as a dependency in `ramda-fantasy`

buzzdecafe: it *is* curried, just done manually

CrossEye:
> are you suggesting using the "constructor" instead of the `fromNullable`, more or less as I used to?

Yes.   Obviously I'm not communicating well if that was still a question.

src/Either.js
TheLudd: ahh case 1? 
I see. 

...:sheep:

buzzdecafe:
@TheLudd my apologies for the non-reply. I suppose we could overload the constructor to work as `fromNullable`, something in that FL thread scared me off of that idea. Any thoughts on that approach @davidchambers ?

TheLudd:
@buzzdecafe NP. So many issues and PRs...
I also noticed your comment in the FL thread, that is why I asked you. Because I never really understood why it was a bad idea.

buzzdecafe:
i will try the overloaded constructors in a separate PR, unless there are any objections. Any other issues with the PR as it stands now?

TheLudd:
:horse: (I hereby claim the horse as my :+1: )

CrossEye:
No other issues.

davidchambers:
Sanctuary provides [`toMaybe`][1] which mirrors [`fromMaybe`][2] which was lifted from [purescript-maybe][3].

This project has less of a PureScript influence, so using the `Maybe` constructor to create `Just`s and `Nothing`s seems reasonable.


[1]: https://github.com/plaid/sanctuary/blob/v0.3.0/index.js#L119-L122
[2]: https://github.com/plaid/sanctuary/blob/v0.3.0/index.js#L107-L117
[3]: https://github.com/purescript/purescript-maybe#values

buzzdecafe:
ok, i'll open a separate PR for that issue

PR #13, opened 2015-02-17
Add equality test for Future

TheLudd:


TheLudd:
Draft of how future equality could work in the test environment.

buzzdecafe:
interesting, and certainly better than what is (barely) there now

TheLudd:
As I understand it, this should be enough to run all the tests in `test/types.js`

buzzdecafe:
does it pass `npm test` now?

TheLudd:
It passes `npm test`
I also added types tests. They pass..

buzzdecafe:
:cow:

TheLudd:
Coolers. Do we also add tests for `map`, `chain` et al? Like

```
var result = R.map(R.add(1), Future.of(1));
assert.equal(true, Future.of(2).equals(result)
```

buzzdecafe:
yeah, i think we should have some way of distinguishing unit tests from "legal compliance" tests.

PR #14, opened 2015-02-17
Add tests for Future .map and .chain

TheLudd:
Simple tests for .map and .chan functions in Future. Also modified the
.equals function to give better error messages in case of failure.


buzzdecafe:
LGTM

PR #18, opened 2015-02-18
Remove run from Reader prototype and make it an instance attribute

TheLudd:
The `run` method on the Reader prototype was only delegating to the
internal function supplied to the constructor. A better solution for
this is to simply name the input function `run`. No need to use
`fn.call(this, arguments)`. Tests still pass.


TheLudd:
This is analogous to the implementation of Futures `fork`

PR #20, opened 2015-02-21
Fix `Reader`.

joneshf:
Don't merge this yet.

So, the tests pass before and after this change, but this is a fairly huge change as far as semantics go. Before, it wasn't a `Functor`, `Apply`, `Chain`, or `Monad` (as all of the functions depended on `chain` working properly), now it "should" be. I say "should" because I'm not actually sure this fix actually fixed it. This tells me that the tests aren't actually testing the semantics here, so we should discuss what should change such that the tests work properly.

I have not looked at the tests yet to see where the failure is. Maybe someone can provide some insight.

### DIFF
--- a/src/Reader.js
+++ b/src/Reader.js
@@ -13,8 +13,8 @@ Reader.run = function(reader) {
 
 Reader.prototype.chain = function(f) {
     var reader = this;
-    return new Reader(function() {
-        return f(reader.run()).run();
+    return new Reader(function(r) {
+        return f(reader.run(r)).run(r);
     });
 };
 


buzzdecafe:
ok, i'll also take a look. i've never been totally convinced by the test suite -- testing for legality is different from testing units of work, but at present the tests are unit tests. so that is an area i would also like to improve, and maybe this is an opportunity to do so.

joneshf:
So, it looks like the tests rely on an `equals` method, like here: https://github.com/ramda/ramda-fantasy/blob/master/test/types.js#L24 But you can't define meaningful equality for functions. Maybe it's time to look into property based testing using something like https://github.com/fantasyland/fantasy-check, https://github.com/robotlolita/claire, or bring in the big guns with https://github.com/purescript/purescript-quickcheck or https://github.com/purescript-contrib/purescript-strongcheck ?

buzzdecafe:
do you have a recommendation?

joneshf:
Well, I've used claire, ps-qc and ps-sc in anger before, but not fantasy-check (though it seems fairly straight forward). Claire is simple and easy to get going. ps-qc is powerful, but comes with a huge change in development. ps-sc is more robust than ps-qc, but again huge change in development.

Were it my project the whole thing would be in ps, and the choice would be ps-sc easily. However, this is a ramda project and I imagine you actually want people to contribute :). So I'd shy away from the ps stuff. It's still not super easy to inject some ps stuff into an arbitrary project (but getting better).

So I guess it comes down to claire vs fantasy-check. Having only experience with claire doesn't make for a fair choice, so I'll swing it back to you. :)

buzzdecafe:
> Were it my project the whole thing would be in ps, and the choice would be ps-sc easily.

understood -- i see ramda as more of a gateway drug for purescript. 

`claire` looks pretty straightforward, so i'll see about integrating that. thanks

buzzdecafe:
i think this change should be merged -- any reason why not?

joneshf:
Oh, sorry. Yes, it's fine to merge this.

PR #22, opened 2015-02-28
Add Maybe getter methods

TheLudd:
This adds the methods `get`, `isNothing` and `isSomething` on `Maybe`.

`get` on a `Just` returns the value inside. Get on a `Nothing` throws an
error.

`isNothing` returns true for `Nothing` instances and false for `Just`
instances. The opposite is true for `isSomething`.

### DIFF
--- a/test/maybe.test.js
+++ b/test/maybe.test.js
@@ -93,3 +93,29 @@ describe('Maybe', function() {
     });
 
 });
+
+describe('Maybe usage', function() {
+
+  it('should allow the user to check if the instance is a Nothing', function() {
+    assert.equal(true, Maybe(null).isNothing());
+    assert.equal(false, Maybe(42).isNothing());
+  });
+
+  it('should allow the user to check if the instance is a Just', function() {
+    assert.equal(true, Maybe(42).isJust());
+    assert.equal(false, Maybe(null).isJust());
+  });
+
+  describe('#getOrElse', function() {
+
+    it('should return the contained value for if the instance is a Just', function() {
+      assert.equal(42, Maybe(42).getOrElse(24));
+    });
+
+    it('should return the input value if the instance is a Nothing', function() {
+      assert.equal(24, Maybe(null).getOrElse(24));
+    });
+
+  });
+
+});
--- a/src/Maybe.js
+++ b/src/Maybe.js
@@ -74,5 +74,21 @@ _Nothing.prototype.equals = function(that) {
     return that === _nothing;
 };
 
+Maybe.prototype.isNothing = function() {
+  return this === _nothing;
+};
+
+Maybe.prototype.isJust = function() {
+  return this instanceof _Just;
+};
+
+_Just.prototype.getOrElse = function(a) {
+  return this.value;
+};
+
+_Nothing.prototype.getOrElse = function(a) {
+  return a;
+};
+
 module.exports = Maybe;
 


buzzdecafe:
isn't the whole point that you *can't* `get` the value? i think if we want to go this way we should stick to the spec and implement `Comonad` which has an `extract` method

TheLudd:
@buzzdecafe I just read the fantasy-land specification and saw that the comonad spec has changed (right?). And I agree that we should change `get` to `extract`

TheLudd:
Looking at the spec again I thing I will remove get from this PR entierly. Replacing it seems to be a bigger job since comonad requires extend and I don't think there are any tests at all yet for those types.

However, what about `isSomething`/`isNothing`?

buzzdecafe:
couldn't you define `isNothing` on `Maybe` prototype:

`Maybe.prototype.isNothing = function() { return this === _nothing; };`

likewise for `isJust`:

`Maybe.prototype.isJust = function() { return this instanceof _Just; };`

?

TheLudd:
Sure, can do that.

You think `isJust` is a better name? I'd say I prefer the word something as it is the opposite of nothing.

joneshf:
You can't implement `Comonad` for `Maybe a`. What does `extract` do on a `Nothing`? You can implement `Extend` though. 

I think a function like `getOrElse` is more worthwhile:

```javascript
Nothing.prototype.getOrElse = function(x) {
  return x;
}
Just.prototype.getOrElse = function(x) {
  return this.value;
}
```

It won't throw an error at runtime.

buzzdecafe:
thanks @joneshf that sounds good to me

davidchambers:
An alternative to `getOrElse`:

```javascript
//  fromMaybe :: a -> m a -> a
Maybe.fromMaybe = curry(function(x, m) {
  return m instanceof Just ? m.value : x;
});
```

buzzdecafe:
i think `isJust` is better because `Maybe a = Nothing | Just a`

joneshf:
> @TheLudd i think isJust is better because Maybe a = Nothing | Just a

:+1:

buzzdecafe:
`getOrElse` sounds somewhat *threatening*. But I think that may be the way to go. That is a Scala thing, yes?

joneshf:
Yeah, it's a scala thing, but `fromMaybe` is fine as well.

buzzdecafe:
the only objection I have to `fromMaybe` is that the type name is in the function, whereas `getOrElse` could work with any type

CrossEye:
> `getOrElse` sounds somewhat _threatening_.

Name for a new library: `bully.js`.

buzzdecafe:
@TheLudd do you want to revise this PR?

TheLudd:
Yes I will. Just going through some crazy times with my company right now so there is little time for other stuff. 

But what is the consensus on the changes to be made? 
 * `isSomething` -> `isJust`
 * add `getOrElse`?

buzzdecafe:
> isSomething -> isJust
> add getOrElse?

:+1:

lmk if you are too busy, and want me to do it. I want to start improving test covrage, so need this stuff to keep going.

TheLudd:
I can do it later today.

TheLudd:
@buzzdecafe :cow2:?

buzzdecafe:
indeed :cow2:

buzzdecafe:
thanks!

joneshf:
:+1:

buzzdecafe:
i have also added `isJust/isNothing` static on `Maybe` to test an arbitrary object. See #25

test/maybe.test.js
```diff
@@ -93,3 +93,29 @@ describe('Maybe', function() {
     });
 
 });
+
+describe('Maybe usage', function() {
+
+  it('should allow the user to check if the instance is a Nothing', function() {
```
joneshf: I feel like there's a law hiding in here. Something like:

```javascript
jsv.assert(jsv.forall(m, function(m1) {
  (m1.isJust() && !m1.isNothing()) || (!m1.isJust() && m1.isNothing());
});
```

joneshf: Oh! It's xor!

buzzdecafe: indeed, it's one or the other but not both. shall i add a test to that effect?

joneshf: Seems to make sense. 

I think there's a law for `getOrElse` also:

```javascript
jsv.assert(jsv.forall(m, 'nat', function(m1, n) {
  if (m1.getOrElse(n) !== n) {
    return m1.isJust();
  } else {
    return m1.isNothing() || m1.equals(Maybe.Just(n));
});
```

I wonder if it can be simplified though.

PR #23, opened 2015-02-28
Add table in README that shows which types implement which interfaces

TheLudd:


TheLudd:
Preview here: https://github.com/TheLudd/ramda-fantasy/tree/type-table

Just thought this was a nice way to display what is in the project. I left out Functor, Apply, Applicable and Chain since: a, They are implied by the Monad interface and b, the table got too wide if I included them =)

davidchambers:
Good idea. I would use "✔︎" rather than "__X__", though.

TheLudd:
@davidchambers Done.

Is the table correct otherwise?

buzzdecafe:
LGTM. do we want Functor, Apply, et al?

TheLudd:
@buzzdecafe 
I put those in at first but the table got too wide (scroll bar). And they are all implied by Monad. The only case we would really need them would be if we had some type that was a Functor but not an Apply or something like that.

buzzdecafe:
> they are all implied by Monad

Fair enough. I hope we don't get into a situation where we have functors that don't implement the other monadic behaviors.

Issue #21, opened 2015-02-25
property testing with claire


buzzdecafe:

buzzdecafe:
superseded by #24

PR #24, opened 2015-03-01
Jsverify

joneshf:
This isn't a complete PR.

Re: #20 I wasn't aware of `jsverify`. It seems like a good choice. Plus @robotlolita endorsed it.

As a benefit of having property based tests, another bug was found.

If you want to go this route I can throw together the rest of the tests sometime this coming week. If not, close this PR.

### DIFF
--- a/src/Maybe.js
+++ b/src/Maybe.js
@@ -41,7 +41,7 @@ _Just.prototype.ap = function(m) {
     return m.map(this.value);
 };
 
-_Nothing.prototype.ap = util.identity;
+_Nothing.prototype.ap = util.returnThis;
 
 // applicative
 // `of` inherited from `Maybe`
@@ -56,7 +56,7 @@ _Just.prototype.chain = util.baseMap;
 _Nothing.prototype.chain = util.returnThis;
 
 
-// 
+//
 _Just.prototype.datatype = _Just;
 
 _Nothing.prototype.datatype = _Nothing;
--- a/package.json
+++ b/package.json
@@ -16,6 +16,7 @@
     "ramda": "^0.10.0"
   },
   "devDependencies": {
+    "jsverify": "^0.5.1",
     "mocha": "^2.1.0",
     "uglify-js": "2.4.x",
     "xyz": "0.5.x"
--- a/test/maybe.test.js
+++ b/test/maybe.test.js
@@ -1,12 +1,43 @@
 var R = require('ramda');
 var assert = require('assert');
 var types = require('./types');
+var jsv = require('jsverify');
 
 var Maybe = require('..').Maybe;
 
+var MaybeGen = R.curry(function(a, n) {
+    return n % 2 == 0 ? Maybe.Just(a.generator(n)) : Maybe.Nothing();
+});
+
+var MaybeShow = R.curry(function(a, m) {
+    if (m instanceof Maybe.Just) {
+        return "Just(" + a.show(m.value) + ")";
+    } else {
+        return "Nothing";
+    }
+});
+
+var MaybeShrink = R.curry(function(a, m) {
+    if (m instanceof Maybe.Just) {
+        return [Maybe.Nothing()].concat(a.shrink(m.value).map(Maybe.Just));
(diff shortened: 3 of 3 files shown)


joneshf:
Whoops, I didn't see #21 before starting this.

buzzdecafe:
no worries, i put #21 up as a reminder for me. since you've already made some progress, i'm happy to go in this direction

davidchambers:
This is exciting!

src/Maybe.js
```diff
@@ -41,7 +41,7 @@ _Just.prototype.ap = function(m) {
     return m.map(this.value);
 };
 
-_Nothing.prototype.ap = util.identity;
+_Nothing.prototype.ap = util.returnThis;
```
buzzdecafe: oops! nice catch

joneshf:
Oh, I still had more to go on here :).

I added some helpers here: https://github.com/joneshf/ramda-fantasy/blob/jsverify/src/Either.js#L30-L36 because I realized that `instanceof` check doesn't actually work. That means that these two checks don't work either: https://github.com/joneshf/ramda-fantasy/blob/jsverify/test/maybe.test.js#L13 https://github.com/joneshf/ramda-fantasy/blob/jsverify/test/maybe.test.js#L21.

Rather than duplicate the effort of #22, I'll just wait until that PR is merged before updating those tests.

I'm not sure there's a nice way to use jsverify to test `Reader` properly with the way tests are now. I've been thinking about changing the functions in `test/types.js` a bit. If they take an equivalence function with how to evaluate each side of the equation, we should be able to make testing work for all things now and in the future:

```javascript
id: function(equiv, obj) {
  return equiv(obj, obj.map(identity));
},
compose: function(equiv, obj, f, g) {
  return equiv(obj.map(function(x) { return f(g(x)); }), obj.map(f).map(g));
}
```

So in the case of `Reader`, we could have:

```javascript
var eqReader = function(x) {
  return function(a, b) {
    return a.run(x) === b.run(x);
  };
};
```

Thoughts?

buzzdecafe:
ya `m instanceof Maybe.Just` won't work, since `Maybe.Just` is a factory and not a constructor. 

This works:

```
Maybe.isJust = function(x) {
    return x instanceof _Just;
};

// Example:
~/dev/ramda-fantasy [master|✚1⚑2] 
18:50 $ node
> Maybe = require('./').Maybe
> j = Maybe.Just(1);
{ value: 1 }
> Maybe.isJust(j)
true
```

So I am surprised the `Either` checks don't.

buzzdecafe:
... and passing in an `eq` function looks like an elegant solution to me

joneshf:
I'm curious about `IO` and `Future`. 

The `IO` tests passing relies on the execution order of the tests. That seems like it's going to get messy if bugs are found, or tests need to change. Also, they seem to be doing too much.

For instance: [this](https://github.com/ramda/ramda-fantasy/blob/3925caee370ae26eb5e6a0f188658e0e057e2a7c/test/io.test.js#L43) depends on [this line](https://github.com/ramda/ramda-fantasy/blob/3925caee370ae26eb5e6a0f188658e0e057e2a7c/test/io.test.js#L42). But all that should be checked in that place is the `Functor` interface. I think it'd be better to test `IO` purely there. And then have unit tests that use that `logger`. Are you okay with that change?

What is `Future`? It seems like it'd have the haskell type:

```haskell
data Future c a b = Future (a -> c) (b -> c)
```

Or is it something else?

buzzdecafe:
>  I think it'd be better to test IO purely there. And then have unit tests that use that logger. Are you okay with that change?

:+1: 

> What is Future? 

good question, my understanding of it is fragile; based on class I took and folktale implementation.

joneshf:
No worries, I'll peruse folktale.

buzzdecafe:
apologies, look here:

https://github.com/DrBoolean/factfiction/blob/306285512c5ab1a282eaf5faee2a4ae6f62d79fc/js/data.future.umd.js

PR #26, opened 2015-03-25
Make Future.ap run in parallel

TheLudd:
This change to `ap` for Futures does not wait for its own fork to be
resolved before forking the input furture. The effect is, among other
things, that `lift` can be used in order to call a method with the
result of several futures that have been resolved in parallel.

### DIFF
--- a/src/Future.js
+++ b/src/Future.js
@@ -1,3 +1,4 @@
+var R = require('ramda');
 module.exports = Future;
 
 // `f` is a function that takes two function arguments: `reject` (failure) and `resolve` (success)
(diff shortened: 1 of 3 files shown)


TheLudd:
I mentioned briefly on gitter that I had made an implementation of `Futute.ap` that could run in parallel. This is it. Open for feedback. I don't find it very pretty and if someone has suggestions for making it prettier then I am listening. I think I would prefer it to work like this however.

TheLudd:
This also affects commute which is done in a series prior to this change but in parallel with this change.

buzzdecafe:
any tests for this? it would be nice to see examples in action

TheLudd:
@buzzdecafe The version you looked at did not work. I have restored my original version wich is the unpretty one. Either way I can supply tests and exmaples

src/Future.js
```diff
@@ -15,7 +15,35 @@ Future.prototype.map = function(f) {
...
+      applyFn = fn;
+      resolveIfDone();
+    });
+
+    m.fork(rejectIfFirst, function(v) {
+      val = v;
+      resolveIfDone();
+    });
+
+  }.bind(this));
```
buzzdecafe: this means >= IE 9 --not a huge deal, but it is pretty easy to keep a reference to original `this` in closure so `bind` is probably not necessary here.

buzzdecafe:
@TheLudd i'm inclined to favor the working one over the pretty one. :smile:

src/Future.js
TheLudd: Did not know about that restriction. I can fix.

TheLudd:
@buzzdecafe Compability fix made and  tests are added

buzzdecafe:
can you squash? then i will merge

TheLudd:
@buzzdecafe squashed

PR #29, opened 2015-04-06
Add bimap to Either

TheLudd:
This adds `bimap` to `Either` which takes two functions an applies the
first one to the left value if existing and the second one to the right
value if existing. Inspired by folktale.


TheLudd:
When I say "inspired by" I mean "copied from".

buzzdecafe:
:cow2:

TheLudd:
Merge? I have reservations against merging my own PRs =)

PR #30, opened 2015-04-06
Add reject to Future

TheLudd:
Complement function to `of`. Creates a future that will be rejected with
the input value.

### DIFF
--- a/src/Future.js
+++ b/src/Future.js
@@ -64,3 +64,9 @@ Future.prototype.chain = function(f) {  // Sorella's:
 // monad
 // A value that implements the Monad specification must also implement the Applicative and Chain specifications.
 // see above.
+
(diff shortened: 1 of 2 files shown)


buzzdecafe:
:cow2:

are there laws we should be concerned about satisfying for `reject`?

TheLudd:
I don't think so since it is completely outside the fantasy-land specification.

CrossEye:
>> are there laws we should be concerned about satisfying for reject?

> I don't think so since it is completely outside the fantasy-land specification.

That might be a _non sequitur_.  We would have to decide on such laws and not leave it to the FanatasyLand team, but that doesn't mean that no such laws should exist.

buzzdecafe:
will merge now, swing back later

PR #31, opened 2015-04-07
Add liftA2 and liftA3

TheLudd:
The lift function specifically implemented for fantasy land compatible
types. Comes in two versions, for functions of airity 2 and 3.

### DIFF
--- a/index.js
+++ b/index.js
@@ -3,6 +3,8 @@ module.exports = {
     Future: require('./src/Future'),
     Identity: require('./src/Identity'),
     IO: require('./src/IO'),
+    lift2: require('./src/lift2'),
+    lift3: require('./src/lift3'),
     Maybe: require('./src/Maybe'),
     Tuple: require('./src/Tuple'),
     Reader: require('./src/Reader')
--- a/src/lift2.js
+++ b/src/lift2.js
@@ -1,0 +1,5 @@
+var R = require('ramda');
+
+module.exports = R.curryN(3, function liftA2(f, a1, a2) {
+  return a1.map(f).ap(a2);
+});
--- a/src/lift3.js
+++ b/src/lift3.js
@@ -1,0 +1,5 @@
+var R = require('ramda');
+
+module.exports = R.curryN(4, function liftA2(f, a1, a2, a3) {
+  return a1.map(f).ap(a2).ap(a3);
+});
--- a/test/lift2.test.js
+++ b/test/lift2.test.js
@@ -1,0 +1,23 @@
+var lift2 = require('../src/lift2');
+var Identity = require('../src/Identity');
+var R = require('ramda');
+var assert = require('assert');
+
+describe('lift2', function() {
+
+  var i1 = Identity.of(1);
+  var i2 = Identity.of(2);
+
+  it('lifts the values of two applys into a curried function', function() {
+    var result = lift2(R.add, i1, i2);
+    assert.equal(true, Identity.of(3).equals(result));
+  });
+
+  it('is itself curried', function() {
+    var step1 = lift2(R.add);
+    var step2 = step1(i1);
+    var result = step2(i2);
+    assert.equal(true, Identity.of(3).equals(result));
+  });
+
+});
--- a/test/lift3.test.js
+++ b/test/lift3.test.js
@@ -1,0 +1,29 @@
+var lift3 = require('../src/lift3');
+var Identity = require('../src/Identity');
+var R = require('ramda');
+var assert = require('assert');
+
+describe('lift3', function() {
+
+  var combine3 = R.curry(function(a, b, c) {
+    return [ a, b, c ].join(' ');
+  })
+
+  var i1 = Identity.of('foo');
+  var i2 = Identity.of('bar');
+  var i3 = Identity.of('baz');
+
+  it('lifts the values of three applys into a curried function', function() {
+    var result = lift3(combine3, i1, i2, i3);
+    assert.equal(true, Identity.of('foo bar baz').equals(result));
+  });
+
+  it('is itself curried', function() {
+    var step1 = lift3(combine3);
+    var step2 = step1(i1);
+    var step3 = step2(i2);
+    var result = step3(i3);
+    assert.equal(true, Identity.of('foo bar baz').equals(result));
+  });
+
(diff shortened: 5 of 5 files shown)


TheLudd:
We discussed whether to have functions like these in ramda-fantasy or not. I would prefer to include `lift` here. Partly because I prefer to use these version over the generic one and partly because the elegant way they are implemented. 

I know that `lift` already exists in ramda but these would be my preferences when working with Apply instances.

src/lift-a3.js
```diff
@@ -0,0 +1,7 @@
+var R = require('ramda');
+
+module.exports = R.curryN(4, function liftA2(f, a1, a2, a3) {
+  return a1.map(f).ap(a2).ap(a3);
+});
+
+
```
davidchambers: :space_invader:

TheLudd: What does that mean? =)

TheLudd: Empty lines?

davidchambers: Yes. They are `:space_invader:`s. ;)

TheLudd: They are shot down

buzzdecafe:
ramda has variadic lift which should cover these cases. do we need these for external compat?

CrossEye:
I don't know.  Should we have `liftM2`, `liftM3`, as well?  I'm not seeing an advantage.

But I haven't been following the direction of the fantasy project very closely.  Is there a goal to make this a very separate stand-alone implementation of the FantasyLand specification with no dependencies on Ramda?  If that's the case, then I could see where this makes sense.

My original thought with this, long ago, and I don't know if @buzzdecafe was thinking the same way, was that an extension like this was something that could easily be bundled with a version of Ramda for those who wanted these additional features.  I never really thought about it standing on its own.

But we have definitely moved away from the bundling idea, and perhaps even the notion of extensions no longer really makes sense.

TheLudd:
@CrossEye I *do* see this as an addon to ramda. Using these types without ramda doesnt make much sense to me. This is also why I haven been shy to import ramda where needed.

The main difference for me (which I have mentioned before so apologies if I am repeating myself) is that you can do `liftA2(f, a1, a2)` in one go with this function. That is not possible with the current implementation of ramdas lift. You need `lift(2, f)(a1, a2)`. Plus, I actually prefer the specific arity as opposed to passing it in as the first argument.

> I don't know. Should we have liftM2, liftM3, as well? I'm not seeing an advantage.

I don't know what these functions implies actually, that they are for monads? As I see it, the functions I implemented requires the input argumets to be of type `Apply`. Which is what I assume the `A` in `liftA` stands for (I did not come up with the name myself). As a `Monad` is also an `Apply`, no specific lift functions for them will be needed.

CrossEye:
I hadn't caught the difference we had between `lift(2, f)(a1, a2)` and `liftA2(f, a1, a2)`.  I agree that the latter is more Ramda-like.  It's also more closely aligned with the recent goals of updating variadic functions into fixed-arity ones.  I'm still only half-committed to that goal, but if we're going to do it, we're going to have to revisit `lift` in any case.

But two things:  First, I think we need to address this in Ramda itself.  `lift`/`liftN` was a nice experiment, but it looks as though it might not be the best bet and a collection of `lift2`, `lift3`, `lift4` would be better.  Second, the name.  I'm pretty sure the Haskell names are based on Applicatives and Monads, which, at least until recently, didn't actually share the obvious relationship.  I think we simply don't need to care about that, and can deal with them in a more generic way and simply call the function `lift`, with the only suffix being the parameter count.

buzzdecafe:
> deal with them in a more generic way and simply call the function lift, with the only suffix being the parameter count.

in which case the existing abstractions of `lift/liftN` are already there. Problem solved!

buzzdecafe:
ping -- how are we going to proceed here?

TheLudd:
Well, I'd prefer to have these versions in. At least until it is decided if ramdas `lift` goes through any change or not. And for that reason I'd also prefer to keep the names with the A at the end to clarify which is which.

buzzdecafe:
i'm inclined to agree with @CrossEye that the `A` or `M` doesn't add much for us. consider the Haskell sigs:

```hs
liftA2 :: Applicative f => (a -> b -> c) -> f a -> f b -> f c
liftM2 :: Monad m => (a1 -> a2 -> r) -> m a1 -> m a2 -> m r
```

for weakly-typed JS, those amount to the *same thing*. But `lift2, 3, 4` would be fine

CrossEye:
Agreed.  My only question is whether we should just do these in Ramda proper or do it in the fantasy project.  I vote for replacing Ramda's `lift` with `lift2/3/4`.

TheLudd:
The reason for the A as I see it is that the implementation is not a generic lift. It is a lift compatible with fantasy land Apply types.

CrossEye:
Woulld not the identical implementation work for Monads?

I'm also curious about a generic implementation that simply tweaks `liftN`:

    var liftN = _curry2(function liftN(arity, fn) {
      var lifted = curryN(arity, fn);
      var lift = curryN(arity, function() {
         return _reduce(ap, map(lifted, arguments[0]), _slice(arguments, 1));
      });
      return arguments.length > 2 ? lift.apply(this, _slice(arguments, 2)) : lift;
    });

Would that cover everything under consideration?

buzzdecafe:
> return arguments.length > 2 ? lift.apply(this, _slice(arguments, 2)) : lift;

pretty much, AFAICT

TheLudd:
> Woulld not the identical implementation work for Monads?

Yes but the lowest common denominator would be Apply. 

But right. If the generic lift is improved I have less need for this.

TheLudd:
@buzzdecafe Can I take ramda/ramda#1058 as we go with this one? But I change the name to `lift2/lift3`?

buzzdecafe:
yes, exactly. i think `lift_` is a better fit here.

TheLudd:
And we are satsfied with 2 and 3?

buzzdecafe:
that should be enough to get us going. if it turns out we need 4 (or more???) it shouldn't be too much trouble to add it.

TheLudd:
Changes done. I also added them to the index file.

PR #32, opened 2015-04-16
Add bimap to future

TheLudd:
Maps the first function over the rejected value of a Future and the
second function over a resolved value of a future.


buzzdecafe:
nice!

PR #33, opened 2015-04-28
Tuples!

DrBoolean:


davidchambers:
Nice! Is tuple commonly used to refer to a pair specifically? Perhaps `Pair` is a more appropriate name?

DrBoolean:
I should have mentioned, the long term plan is to add sugar for left associative arbitrary length tuples:
`Tuple(x,y,z)  ==  Tuple(Tuple(x,y), z)`

That enables derived instances and keeps things sane. So hopefully soon it'll be more flexible than a Pair.

davidchambers:
In that case, :thumbsup: from me!

src/Tuple.js
```diff
@@ -0,0 +1,74 @@
...
+    xs.map(function(x){
+      if(typeof x.concat != "function") {
+          var display = x.show ? x.show() : x;
+          throw new TypeError(display + " must be a semigroup to perform this operation");
+      };
+    })
+}
+
+Tuple.of = function(x) {
+    return Tuple(x, x);
```
buzzdecafe: is `(x, x)` better than `(x, empty)`?

src/Tuple.js
```diff
@@ -0,0 +1,74 @@
...
+};
+
+function _Tuple(x, y) {
+    this[0] = x;
+    this[1] = y;
+    this.length = 2;
+};
+
+function ensureConcat() {
+    var xs = [].slice.call(arguments)
```
buzzdecafe: since this is a privateh helper anyways, maybe we should just pass a list, rather than arg-slicing

src/Tuple.js
DrBoolean: It should be `(empty, x)` since `snd` is the functor target, but that requires `Monoid a => Tuple a b` which is only necessary for the applicative instance.

src/Tuple.js
DrBoolean: Good call! I'll update that

src/Tuple.js
```diff
@@ -0,0 +1,74 @@
...
+    xs.map(function(x){
+      if(typeof x.concat != "function") {
+          var display = x.show ? x.show() : x;
+          throw new TypeError(display + " must be a semigroup to perform this operation");
+      };
+    })
```
buzzdecafe: this `map` for side-effects feels a bit strange, but i wonder if making `ensureConcat` return a Bool if it is concatable is any better. I don't think so...

buzzdecafe:
So Dr. Boolean! I was in your Hardcore Functional Programming in JS class -- but now it seems the student has become the master! Mu-wahaha!

j/k of course, very glad to see this PR :smile:

DrBoolean:
Haha. Glad I can contribute!

src/Tuple.js
```diff
@@ -0,0 +1,73 @@
...
+    return x[1];
+};
+
+_Tuple.prototype.of = function(x) {
+    return Tuple(this[0], x)
+}
+
+// semigroup
+_Tuple.prototype.concat = function(x) {
+    ensureConcat([this[0], this[1], x[0], x[1]])
```
CrossEye: We're not testing that `this[0]` and `x[0]` have the same type, and would probably never get to that point.  So is there any good reason to even pass `x[0]` (or of course `x[1]`?)  It seems that we only need to check that the components of `this` can call `concat`.

DrBoolean: Right! It's really only `this[0] / this[1]` that need `concat()`. I'll adjust.

src/Tuple.js
```diff
@@ -0,0 +1,73 @@
...
+
+// apply
+_Tuple.prototype.ap = function(m) {
+    ensureConcat([this[0]])
+    return Tuple(this[0].concat(m[0]), this[1](m[1]));
+};
+
+// setoid
+_Tuple.prototype.equals = function(that) {
+    return that instanceof _Tuple && this[0] === that[0] && this[1] === that[1]
```
CrossEye: This will only work with primitives.  Do we want to use something like Ramda's `eqDeep` ?

DrBoolean: Sure do! I was sticking with the pattern set by the other objects. Perhaps all should be updated to use that?

buzzdecafe: yes, definitely. i'll make an issue for it

CrossEye:
I'm a little confused about underlying motivation.  Why is `snd` the functor target?  Is this simply a way for Tuple to be the underlying implementation of Either?  Or is there some more fundamental reason for it?  Similarly, why left-associative?  There would be an interesting (although probably far from efficient!) way to work with lists if these were right-associative.  Is there prior art that dictates these decisions?

DrBoolean:
Mainly I took the standard tuple instance from haskell and left associativity from purescript.

The functor instances is useful for carrying information along like a simplified reader:
```
Tuple({my: "config"}, [1,2,3]).map(reverse)
//=> Tuple({my: "config"}, [3,2,1])
```

This differs quite a bit from Either as a pure error handling technique.

It's also useful for forking a program with things like `of` and `bimap`
```
compose(bimap(add(1), add(2)), Tuple.of)(3)
//=> Tuple(4,5)
```

And really useful for sorted key/value pairs.

The left associatively apparently allows one to derive instances for free from tuples of arbitrary lengths, but i still have to investigate why.

buzzdecafe:
this LGTM -- can you please squash to 1 commit?

DrBoolean:
squarshed

CrossEye:
:herb:

buzzdecafe:
nice! i'll fix up the equality thing across all types in a separate PR

PR #40, opened 2015-05-16
bowerize. fixes #39

buzzdecafe:

PR #41, opened 2015-05-16
automate release process

davidchambers:
In light of #40, it's particularly important that we automate the publishing process: there are several steps one might forget when publishing manually.

You might prefer to add the commands as npm scripts, @buzzdecafe. I'll open another pull request for comparison, if you're interested.


buzzdecafe:
if you're offering, then sure, i'd like to see the npm version. Although I can guarantee I will *always* have `Make` available, not so sure about everybody else.

buzzdecafe:
superseded by #42

PR #44, opened 2015-05-22
Extend for Either

gilligan:
Extend is trivial to implement for Either (as discussed on gitter.im with @joneshf )


buzzdecafe:
thanks! i gotta add some tests around this stuff

PR #45, opened 2015-05-22
fixed a typo

ChetHarrison:
Not a big deal but I think the we had a little copy/past action here. liftA3 was called liftA2.


buzzdecafe:
should probably just be `lift3`, since it ought to work with M or A

TheLudd:
@ChetHarrison Would you like to update your PR to fix this? Change the internal names to `lift2` and `lift3`?

TheLudd:
Solved by #65

PR #47, opened 2015-05-23
Loosen up timeout for Future#ap parallel test

gilligan:
The test always fails for me because it apparently just won't finish within the 20ms timeout. Raising it to 25ms fixed it for me and since 2*15 > 25 this stills proofs non-sequential execution.


TheLudd:
If there are better ways to prove parallelism I am all ears.

PR #48, opened 2015-06-03
style: use two spaces for indentation

davidchambers:
The code currently uses two spaces in some places and four spaces in others.


buzzdecafe:
yes, it's a bit messy. this is great, thx

PR #50, opened 2015-06-03
define "toString" method for each constructor

davidchambers:
This makes ramda-fantasy types compatible with [`R.toString`][1].

Should I update this pull request to remove the `show` method from __Tuple.js__?

[1]: http://ramdajs.com/docs/#toString

### DIFF
--- a/src/Maybe.js
+++ b/src/Maybe.js
@@ -1,3 +1,5 @@
+var R = require('ramda');
+
 var util = require('./internal/util.js');
 
 function Maybe(x) {
@@ -93,6 +95,14 @@ _Just.prototype.getOrElse = function() {
 
 _Nothing.prototype.getOrElse = function(a) {
   return a;
+};
+
+_Just.prototype.toString = function() {
+  return 'Maybe.Just(' + R.toString(this.value) + ')';
+};
+
+_Nothing.prototype.toString = function() {
+  return 'Maybe.Nothing()';
 };
 
 module.exports = Maybe;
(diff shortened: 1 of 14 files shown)


src/Maybe.js
```diff
@@ -95,4 +97,12 @@ _Nothing.prototype.getOrElse = function(a) {
   return a;
 };
 
+_Just.prototype.toString = function() {
+  return 'Maybe.Just(' + R.toString(this.value) + ')';
+};
+
+_Nothing.prototype.toString = function() {
+  return 'Maybe.Nothing()';
```
buzzdecafe: are the parens necessary here? i'm treating `Nothing` a bit differently than in Sanctuary

buzzdecafe:
> Should I update this pull request to remove the show method from Tuple.js?

sure!

src/Maybe.js
davidchambers: The idea is that `eval(R.toString(x))` should be equivalent to `x` for every `x`.

It's up to you, though. I'll remove the parens if you'd prefer the string representation to omit them.

davidchambers:
I've removed the `show` method.

src/Maybe.js
buzzdecafe: no strong feeling. i suppose the fact that there is only one `Nothing` in ramda-fantasy is an implementation detail

buzzdecafe:
good stuff, thx

PR #53, opened 2015-06-17
remove applicative from Tuple

### LINKED ISSUE #51: Tuple Applicative Behaviour
As briefly discussed on Gitter, the current Applicative implementation for `Tuple` doesn't behave.

```js
Tuple.of(R.identity).ap(Tuple.of({}));
//=> [TypeError: function f1(a) { ... } be a semigroup to perform this operation]
```

I currently see three options here:
- (1) Downgrade Tuple to `Apply` by dropping `of`.

- (2) Change the current `ap` implementation to:
```js
_Tuple.prototype.ap = function(m) {
  return Tuple(this[0](m[0]), this[1](m[1]));
};
```

- (3) Change the current `of` implementation to something like:
```js
Tuple.of = function(x) {
  return Tuple({ concat: R.identity }, x);
};
```

I think I'd like to see `Tuple` take on option 2 and if desirable, introducing a `Writer` type that can implement option 3, but I'm keen to hear the thoughts of others.

buzzdecafe:
fixes #51


test/tuple.test.js
```diff
@@ -107,17 +95,17 @@ describe('Tuple usage', function() {
...
+//      var tpl = Tuple.of('pillow pets');
+//      assert.equal('pillow pets', tpl[0]);
+//      assert.equal('pillow pets', tpl[1]);
+//    });
+
+//    it('should maintain the current fst if it already has one', function() {
+//      var tpl = Tuple.of(100).of('buckaroonies');
+//      assert.equal(100, tpl[0]);
+//      assert.equal('buckaroonies', tpl[1]);
+//    });
```
davidchambers: Why not remove these?

buzzdecafe: i don't recall now. so i guess i should

PR #54, opened 2015-06-18
Adding chainReject for chaining operations off of failure cases. 

jethrolarson:
e.g. If you make a ajax request and you want to process error data differently than success data.

This is Task::orElse in folktale's data.task

### DIFF
--- a/src/Future.js
+++ b/src/Future.js
@@ -53,10 +53,22 @@ Future.prototype.of = Future.of;
 //  f must be a function which returns a value
 //  f must return a value of the same Chain
 //  chain must return a value of the same Chain
+//:: Future a, b => (b -> Future c) -> Future c
 Future.prototype.chain = function(f) {  // Sorella's:
   return new Future(function(reject, resolve) {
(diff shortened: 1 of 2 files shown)


TheLudd:
Just thinking about the name here. A similar method could be added to `Either`.  Would it be a good idea to give them the same name? Same goes for mapping over the rejected or left value. Is there any standard terminology for this?

jethrolarson:
There was some discussion on gitter. Switching all of Future terms to be
Left/Right is a thought. I'm not sure it makes it more learnable but it
would help the APIs line up. I expect most people learning Future will come
from Promises not Either

jethrolarson:
Also chaining over left isn't part of monad or chain specification so there
might not be an established word. We should probably impl bichain and
bitmap for both of these though

buzzdecafe:
i followed the conversation on gitter, and apart from the name i think this is a nice addition. We can either pull as is and revise the name later (my choice), or search for the ideal name and only merge when we've found it.

jethrolarson:
Lib is still alpha, so merge is cheap

PR #55, opened 2015-06-25
Adds Reader Transformer

scott-christopher:
Not sure if there is interest in going down the transformer path or not, but I ended up with this after wanting a Reader/Future combo.

An example:
```js
var Reader = require('./src/Reader'),
    Future = require('./src/Future'),
    ReaderTFuture = Reader.T(Future);

var log = console.log.bind(console);

var delay = d => Future((_, done) => setTimeout(done, d));

var delayedText = ReaderTFuture.ask.chain(env => 
  ReaderTFuture.lift(delay(env.delay).map(() => env.text))
);

delayedText.run({ delay: 1000, text: 'Hello!' }).fork(log, log);
```

If desirable, we could also implement `Reader` via `Reader.T(Identity)`.

I'm not sure how useful this is in the JS space, so I would appreciate comments by anyone who has experience with using `ReaderT`s in general.

### DIFF
--- a/src/Reader.js
+++ b/src/Reader.js
@@ -38,9 +38,7 @@ Reader.prototype.of = function(a) {
 };
 Reader.of = Reader.prototype.of;
 
-Reader.ask = Reader(function(a) {
-  return a;
-});
+Reader.ask = Reader(R.always);
 
 Reader.prototype.equals = function(that) {
   return this === that ||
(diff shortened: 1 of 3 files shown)


src/Reader.js
```diff
@@ -52,4 +52,64 @@ Reader.prototype.toString = function() {
...
+      return new ReaderT(run);
+    }
+    this.run = run;
+  };
+
+  ReaderT.lift = function(m) {
+    return ReaderT(function() {
+      return m;
+    });
+  };
```
davidchambers: This could be defined `R.compose(ReaderT, R.always)`. :)

src/Reader.js
```diff
@@ -52,4 +52,64 @@ Reader.prototype.toString = function() {
...
+
+  ReaderT.ask = ReaderT(function(e) {
+    return M.of(e);
+  });
```
davidchambers: Could this be defined `ReaderT(M.of)`?

davidchambers:
I should read about ReaderT. Looks interesting!

scott-christopher:
I've updated the PR with the suggested tweaks:
```js
ReaderT.lift = R.compose(ReaderT, R.always);
ReaderT.ask = ReaderT(M.of);
```

src/Reader.js
```diff
@@ -52,4 +50,58 @@ Reader.prototype.toString = function() {
...
+      return m.chain(function(a) {
+        return f(a).run(e);
+      });
+    });
+  };
+
+  ReaderT.prototype.map = function(f) {
+    return this.chain(function(a) {
+      return ReaderT.of(f(a));
+    });
```
davidchambers: Could this be written `this.chain(R.compose(ReaderT.of, f))`?

joneshf:
> I'm not sure how useful this is in the JS space, so I would appreciate comments by anyone who has experience with using ReaderTs in general.

I'd venture to say that `Reader` doesn't really offer anything worthwhile (aside from avoiding extending a built in object), since `Reader = Function`, but that `ReaderT` is the worthwhile abstraction.

> I should read about ReaderT. Looks interesting!

It's the Kleisli composition you implemented the other day, just codified into an object.

buzzdecafe:
sorry for the delay, i've been offline lately. was nice. this LGTM :cow2:

TheLudd:
I just tried this out. I am new to the concept but this seems very useful. But I do have some comments on the PR.

 1. There are no tests for `ask` and `lift`. Could these be added? (I did notice that `lift` is tested indirectly, don't know if this is considered enough)
 2. When using `ask` you do `ReaderTFuture.ask.chain`. Is this correct? For the normal reader I have only seen `Reader.ask.map`. But perhaps the transformer requires `chain` instead of `map`?

TheLudd:
Never mind question 2. I now realized that chain is necessary since the inner function returns a `ReaderTFuture`

src/Reader.js
```diff
@@ -52,4 +50,58 @@ Reader.prototype.toString = function() {
...
+  };
+
+  ReaderT.prototype.ap = function(a) {
+    var reader = this;
```
TheLudd: I'd name this variable `readerT`

TheLudd:
@scott-christopher 
Can you rebase and have a look at the remaining comments? After that I am all for a merge

scott-christopher:
@TheLudd I've added a simple test that exercises both `ask` and `lift` (`lift` should also be tested via the `transformer` tests in `types.js`), updated the `reader` variable to `readerT` and rebased.

We should be able to replace the `Reader` instance with `Reader.T(Identity)` too, but that can come later if desired.

TheLudd:
@scott-christopher 

> We should be able to replace the Reader instance with Reader.T(Identity) too, but that can come later if desired.

Yes let's do that in a separate commit but it is a very nice idea to avoid duplicate code.

TheLudd:
Anyway :horse: from me.

PR #56, opened 2015-06-25
Adds Either.either

scott-christopher:


TheLudd:
Where does fold come from? It is not a part of the fantasy-land spec right?

scott-christopher:
I believe the correct term for it is a catamorphism (please correct me if I'm wrong), which is a generalisation of a fold.

```
(Either a b).fold :: (a -> c) -> (b -> c) -> c
```

Haskell has an `either` function which is specialised for `Either` types. Another is `maybe` for `Maybe` types. Scalaz also has `(A \/ B).fold`.

Perhaps another name is preferable to prevent confusion with `Foldable` types?

scott-christopher:
And sorry, it's not part of the FL spec that I'm aware of.

davidchambers:
Could these `fold` methods be used in an abstraction that applies to other data types? If not, it would seem more natural to me to define `Either.either :: (a -> c) -> (b -> c) -> Either a b -> c`. Prototypal inheritance is useful for the general operations, but for anything specific to a data type I'd prefer a regular function.

scott-christopher:
Well a catamorphism (to the best of my limited knowledge) applies functions to each of the data constructors of a type, reducing it down to some other type.

e.g.:
```hs
bool   :: a -> a -> Bool -> a                       -- data Bool = True | False
maybe  :: b -> (a -> b) -> Maybe a -> b             -- data Maybe a = Nothing | Just a
either :: (a -> c) -> (b -> c) -> Either a b -> c   -- data Either a b = Left a | Right b
```

I do agree that it probably would be better to have separate function names for each data type, as the signature will differ for each making it troublesome to use generically across types. I'll update the PR, swapping it from a `fold` method to an `either` function.

scott-christopher:
I've updated the PR to implement this via a static `Either.either` function instead.

The naming `Either.either` feels a little clumsy, but I don't have many other suggestions. Perhaps `Either.choose`, or `Either.choice`?

davidchambers:
I don't have a solution to the naming problem. I agree that `Either.either` is clumsy, but it has the advantage of not relying on its namespace to convey information: `either(a, b, c)` is more descriptive than `choose(a, b, c)`.

scott-christopher:
Yeah, good point. I'm not too hung up on the name, I was just throwing it out there in case someone had any other creative names.

TheLudd:
I think I'd prefer bringing it back to `fold` where each sub type implements their own version of it. 

```
Left.prototype.fold = function(f, _) {
  return f(this.val)
}
Right.prototype.fold = function(_, g) {
  return g(this.val)
}  
```
We can then define `fold` as a function not connected to `Either`:
```
function fold(f, g, a) {
  return a.fold(f, g)
}
```

And other types can implement fold in their own way. `Maybe` could have one to. Whether to fold a `Nothing` with a value or a function has been discussed with some disagreement but I don't see any problem in code like this:
```
var person = { ... }
var maybeAge = maybeProp('age', person)
var canBuyAlcohol = fold(R.always(false), R.gte(R.__, 21), maybeAge)
```

scott-christopher:
The problem in implementing `fold` in a generic fashion for `Either` is that it requires two arguments which doesn't apply to all foldable types.

Given the introduction of the `foldable` FL type, I'd suggest just keeping this as `Either.either` for now to reduce potential confusion.

scott-christopher:
Just to bring some life back to this, is there any concern with introducing this function, or is it just down to picking a name? If it's just a naming issue, can I suggest another option of `Either.choice`?

davidchambers:
:deciduous_tree:

`Either.either` is fine.

On an unrelated note, I'd love to see ramda-fantasy and Sanctuary converge at some point.

scott-christopher:
> On an unrelated note, I'd love to see ramda-fantasy and Sanctuary converge at some point.

I'm curious to hear your thoughts. :)

davidchambers:
I would like Sanctuary to provide all the types ramda-fantasy currently does. Once this is the case, Sanctuary will essentially be a superset of ramda-fantasy. The question for ramda-fantasy contributors will then be whether it's preferable for the library to provide types only, or to additionally provide a collection of functions which take advantage of these types. Were we to choose the latter, ramda-fantasy's scope would then grow to match Sanctuary's. At that point the two libraries would have the same goals and philosophies, so it would be inefficient to develop them in parallel.

PR #57, opened 2015-06-25
Adds Fn type

scott-christopher:
I don't quite know where this sits on a scale from zero to absurd, but it's effectively a newtype wrapper for functions, providing implementations for a number of the fantasy land specs without polluting the global `Function.prototype`.

The result of calling `Fn` is a curried function, with a bunch of methods attached:
```js
var add = Fn((a, b) => a + b);
var mul = Fn((a, b) => a * b);

add(1, 2); // 3
var incr = add(1);
incr(2); // 3

add(2).compose(mul(4))(10); // 42
add(2).map(mul(4))(10); // 48

Fn(R.prepend).ap(R.of)('x'); // [ 'x', 'x' ]

mul(2).chain(a =>
  add(10).chain(b =>
    Fn.of(a + b)))(3) // 19
```

It also adds some pretty printing, which *should* survive currying and also include any partially applied arguments:
```js
var add = function Fn(add(a, b) { return a + b; });
add.toString(); // Fn[add](_, _)
add(1).toString(); // Fn[add](1, _)

var anonId = Fn(x => x);
Fn(anonId).toString(); // Fn(_)
```

If anyone is familiar with the various class implementations for function types, please provide some feedback as this was a bit of a learning exercise for me. In particular, some confirmation that the laws are in fact tested correctly would be most welcome. Unfortunately the existing law assertions in `tests/types.js` couldn't be used because they require the type to implement `equals`, which can't easily be defined for functions. I do however have another change I can submit for allowing a custom equality function to be passed in to the type assertions, which would resolve this.

### DIFF
--- a/test/fn.test.js
+++ b/test/fn.test.js
@@ -1,0 +1,210 @@
+var assert = require('assert');
+
+var Fn = require('..').Fn;
+
+describe('Fn', function () {
+
+  var id = Fn(function (x) {
+    return x;
+  });
+  var add = Fn(function (a, b) {
+    return a + b;
+  });
+  var mul = Fn(function (a, b) {
+    return a * b;
+  });
+  var composeFn = Fn(function (f, g, x) {
+    return f(g(x));
+  });
+  var apply = Fn(function (x, f) {
+    return f(x);
+  });
+
+  it('produces a curried function', function () {
+    var incr = add(1);
+    assert.strictEqual(incr(10), 11);
+  });
+
+  describe('is a Monoid', function () {
+    function Mul(v) {
+      this.val = v;
+    }
+
+    Mul.prototype.empty = Mul.empty = function empty() {
+      return new Mul(1);
+    };
+    Mul.prototype.concat = function concat(b) {
+      return new Mul(this.val * b.val);
+    };
+    Mul.prototype.equals = function equals(b) {
+      return b instanceof Mul && this.val === b.val;
+    };
+
+    var m = new Mul(42);
+    var addMul1 = Fn(function add1(x) {
+      return new Mul(x.val + 1);
+    });
+    var mulMul10 = Fn(function mul10(x) {
+      return new Mul(x.val * 10);
+    });
+
+    it('satisfies left neutral', function () {
+      assert.strictEqual(Fn.empty().concat(id)(m).equals(m), true);
+    });
+
+    it('satisfies right neutral', function () {
+      assert.strictEqual(id.concat(Fn.empty())(m).equals(m), true);
+    });
+
+    it('satisfies associativity', function () {
+      var monoidLaw3L = id.concat(addMul1.concat(mulMul10));
+      var monoidLaw3R = id.concat(addMul1).concat(mulMul10);
+      assert.strictEqual(monoidLaw3L(m).equals(monoidLaw3R(m)), true);
+    });
+  });
+
+  describe('is a Functor', function () {
+    it('satisfies identity', function () {
+      var x = 42, y = 1;
+      assert.strictEqual(add(x).map(id)(y), id(add(x))(y));
+    });
+
+    it('satisfies composition', function () {
+      var x = 42, y = 2, z = 1;
+      assert.strictEqual(
+        id.map(add(x).compose(mul(y)))(z),
+        id.map(mul(y)).map(add(x))(z)
+      );
+    });
+  });
+
+  describe('is an Applicative', function () {
+    it('satisfies identity', function () {
+      var x = 42, y = 1;
+      assert.strictEqual(
+        Fn.of(function (z) {
+          return z;
+        }).ap(add(x))(y),
+        add(x)(y)
+      );
+    });
+
+    it('satisfies composition', function () {
+      var x = 42;
+      assert.strictEqual(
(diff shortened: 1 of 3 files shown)


test/fn.test.js
```diff
@@ -0,0 +1,210 @@
...
+      assert.strictEqual(Fn(function add(a, b) { return a + b; }).toString(), 'Fn[add](_, _)');
+    });
+
+    it('returns the string representation of an anonymous function with args', function () {
+      assert.strictEqual(Fn(function(a, b) { return a + b; }).toString(), 'Fn[?](_, _)');
+    });
+
+    it('returns the string representation of a partially applied named function', function () {
+      assert.strictEqual(Fn(function add(a, b) { return a + b; })(1).toString(), 'Fn[add](1, _)');
+    });
```
davidchambers: This is a nice string representation!

CrossEye:
The `concat` and `empty` look a bit strange to me.  Is there a good reason this type _should_ implement `Semigroup` and `Monoid`?  Other than that, this is quite nice.  In fact, I might have some some real uses for something like this in a side project.

Here are those functions, recast for discussion into ES6:

    Fn.prototype.concat = Fn((g, x) => this(x).concat(g(x)));
    Fn.prototype.empty = Fn.empty = Fn(function (x) => x.empty());

Assume our instance, `f` has type `:: a -> b` and the `g` we supply above is `:: c -> d`, then, since `this(x)` is eventually calling `f(x)`, we get `a == c`.  We also can notice that `b` must have a `concat` function on it, so `b` is a `Semigroup`, and thus `d` must also be a member of that same `Semigroup`.  Thus, if we want `Fn` to be a `Semigroup`, it can only accept functions from a fixed type to a specific `Semigroup`.  That seems very restrictive, and for little gain.  Calling `concat` on functions doesn't seem to add much.

Or am I simply missing some important goodness?

scott-christopher:
If we think of `concat` as having a type of `a -> a -> a`, then a semigroup of these functions translates to:

```js
concat :: (a -> b) -> (a -> b) -> (a -> b)
```

In order to combine the results of the functions being `concat`'d, `b` must also have the constraint of being a `semigroup`. Similarly with `empty`.

For a point of reference, this is the `Monoid` instance for functions in Haskell:
```haskell
instance Monoid b => Monoid (a -> b) where
        mempty _ = mempty
        mappend f g x = f x `mappend` g x
```

joneshf:
`Fn === Reader` :)

scott-christopher:
> Fn === Reader

Err... so it is. :smile: I don't know how I failed to see that, given I've been playing around with the ReaderT recently.

Interestingly, `Fn.ap` appears to be closely related to `R.converge` too.

```js
var add = Fn((a,b) => a + b)
var mul = Fn((a,b) => a * b)
var sub = Fn((a,b) => a - b)

Fn.of(add).ap(mul(2)).ap(sub(5))(2) // 7
R.converge(add, mul(2), sub(5))(2)  // 7
```

I have however just realised that this approach will also always report the length of the decorated functions as `0`, unless a wrapper function of matching arity (e.g. `f0()`, `f1(a)`, `f2(a,b)`, `fN(a,b,...)`, etc) is returned instead :(

CrossEye:
> If we think of `concat` as having a type of `a -> a -> a`, then a semigroup of these functions translates to:

> ```js
> concat :: (a -> b) -> (a -> b) -> (a -> b)
> ```

Yes, you just showed from first principles what I discovered from your implementation.

> For a point of reference, this is the `Monoid` instance for functions in Haskell:
> ```haskell
> instance Monoid b => Monoid (a -> b) where
>         mempty _ = mempty
>         mappend f g x = f x `mappend` g x
> ```

Yes, this makes perfect sense to me in Haskell.  I've been having the same problem with `Monoids` and `Semigroups` in Javascript for many of the ADTs.  It's straightforward to me, to talk about a `Functor`, for instance on a plain `Maybe` type.  And that's what most implementations I've seen in Javascript for `Maybe` are, just a plain type, not `Maybe Int` or `Maybe String` or `Maybe BankAccount`.  But `Functors`, `Applicatives`, and the whole `Monad` hierarchy makes sense here.

However I don't think `Semigroup` (or hence, `Monoid`) really makes sense for these bare types.  Something simply seems to be missing.  And I don't see that defining a `Monoid` on them usually gains one much in any case.  So I simply haven't seen the point.

In any case, it wasn't an objection, just a comment.  Since, as @joneshf pointed out, you're really reimplementing a well-known type, it might be most interesting to see how this compares to the `Reader` already in r-f.  But I'm not doing it.  I'm going to bed, and leaving for vacation first thing in the morning!

Have a great week!

joneshf:
> Interestingly, Fn.ap appears to be closely related to R.converge too.

Yep! They're the same, modulo variadic-ness.

scott-christopher:
>However I don't think Semigroup (or hence, Monoid) really makes sense for these bare types.

Yeah, I don't really know how much any of this makes sense. :) I started out just wanting to play with generalising composition and it grew from there.

> I'm going to bed, and leaving for vacation first thing in the morning!

Have a great vacation!

TheLudd:
@scott-christopher Is this still wanted or was it concluded that `Reader` handles the use cases of this one?

scott-christopher:
My bad. I thought I closed this already.

PR #58, opened 2015-07-02
Adds reduce, maybe and toArray to Maybe

scott-christopher:
Example:
```js
// Maybe.maybe :: b -> (a -> b) -> Maybe a -> b
Maybe.maybe(42, x => 2 * x, Just(10));  // 20
Maybe.maybe(42, x => 2 * x, Nothing()); // 42

// Maybe.toArray :: Maybe a -> [a]
Maybe.toArray(Just(42));  // [42]
Maybe.toArray(Nothing()); // []

// Maybe m => m.reduce :: (z -> a -> z) -> z -> z
Just(32).reduce(R.add, 10);  // 42
Nothing().reduce(R.add, 10); // 10
```

### DIFF
--- a/src/Maybe.js
+++ b/src/Maybe.js
@@ -36,6 +36,12 @@ Maybe.isNothing = function(x) {
   return x === _nothing;
 };
 
+Maybe.maybe = R.curry(function(nothingVal, justFn, m) {
+  return m.reduce(function(_, x) {
+    return justFn(x);
+  }, nothingVal);
+});
+
 // functor
 _Just.prototype.map = function(f) {
   return this.of(f(this.value));
@@ -95,6 +101,14 @@ _Just.prototype.getOrElse = function() {
 
 _Nothing.prototype.getOrElse = function(a) {
   return a;
+};
+
+_Just.prototype.reduce = function(f, x) {
+  return f(x, this.value);
+};
+
+_Nothing.prototype.reduce = function(f, x) {
+  return x;
 };
 
 _Just.prototype.toString = function() {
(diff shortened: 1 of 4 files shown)


buzzdecafe:
:cow2: good stuff

TheLudd:
LGTM as long as the `Nothing` cases are correctly implemented. I find this hard to judge sometimes. 

But looks as if we need to add `Foldable` to the table in the readme. Could you do that as well @scott-christopher?

davidchambers:
:thumbsup:

src/Maybe.js
```diff
@@ -36,6 +36,18 @@ Maybe.isNothing = function(x) {
   return x === _nothing;
 };
 
+Maybe.toArray = function(m) {
```
joneshf: You should be able to derive this once and for all for any `Foldable` as:

```js
function toArray(x) {
  return x.reduce(R.flip(R.append), []);
}
```

scott-christopher: I thought about that, but wasn't sure where it should go.

Is it worth introducing a separate file that can hold functions that apply to `Foldable` instances? It seems a bit too fantasyland focused to end up in Ramda proper.

buzzdecafe: a separate file would be fine, or we have a grab-bag of utility functions here: https://github.com/ramda/ramda-fantasy/blob/master/src/internal/util.js

scott-christopher: I've removed `toArray` from this PR and I'll raise a new PR in a few days time to include a separate `Foldable.js` with a handful of useful functions that operate of foldable types.

scott-christopher:
@TheLudd, I've updated the PR to include a Foldable column to the README.

PR #59, opened 2015-07-07
Catch errors thrown in the fork function

TheLudd:
When futures are forked, if the fork function throws a synchronous error
it is now caught and passed to the reject function.


buzzdecafe:
lgtm

TheLudd:
@buzzdecafe merge?

PR #62, opened 2015-08-11
Add Future.Memoize

TheLudd:
Memoize wraps a `Future` and returns a new `Future` that guarantees
that the inner `Future` will only be forked once. Later forkings will
callback with the cached resolved value or error. If the memoized
`Future` is forked several times before it has been resolved, all
callbacks will be fired once the asynchronous operation is completed.

Apart from feedback on the code, I am open for a discussion on whether this should be a part of ramda-fantasy, or not and if the placement of the function is good or if it should have its own separate file.

### DIFF
--- a/src/Future.js
+++ b/src/Future.js
@@ -105,4 +105,42 @@ Future.prototype.toString = function() {
   return 'Future(' + R.toString(this._fork) + ')';
 };
 
+Future.memoize = function(f) {
+  var status = 'IDLE';
+  var listeners = [];
+  var cachedValue;
+
+  var handleCompletion = R.curry(function(newStatus, cb, val) {
+    status = newStatus;
+    cachedValue = val;
+    cb(val);
+    R.forEach(function(listener) {
+      listener[status](cachedValue);
+    }, listeners);
+  });
+
+  function addListeners(reject, resolve) {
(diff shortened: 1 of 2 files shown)


TheLudd:
There is also a possibility of adding this to the future prototype so that one can do `future.memoize()`. I don't know if it is preferable.

buzzdecafe:
makes more sense to me for it to be static as you have it, than to make it an instance method

scott-christopher:
The implementation looks fine to me. I've got some ideas around Future that may make this unnecessary, but they're still a little way from being something demonstrable.

I also agree that it would be preferable to keep this off the prototype. Perhaps it could be a useful rule of thumb to limit the prototype for fantasy-land methods only?

TheLudd:
Keep it like this and merge then? =)

TheLudd:
@scott-christopher I am interested in you ideas. Just yesterday I started using `Future` in my production enviroment and I felt I needed this, but after some work I am not totally sure. But I would still like to merge this and have it available. Later on it could be removed if there are other better ways to handle previously loaded data.

scott-christopher:
I'll keep you posted if my idea ever materialises into anything, but this is good for now :)

Thanks @TheLudd

PR #64, opened 2015-08-12
Removes `equals` from `IO` and `Reader`.

### LINKED ISSUE #63: IO tests fail with ramda 0.17
I was trying to upgrade to ramda `0.17.x` but the tests for `IO` fail. I can't really see why. Does anyone else have an idea?

scott-christopher:
- `types.js` tests now accept a custom equality function to be passed in.
- Bumps ramda version to 0.17
- [fix] `IO.chain` has been updated to pass on `arguments`

Fixes #63

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "test": "mocha"
   },
   "dependencies": {
-    "ramda": "^0.15.0"
+    "ramda": ">=0.15.0 <0.18.0"
   },
   "devDependencies": {
     "jscs": "1.13.x",
--- a/src/IO.js
+++ b/src/IO.js
@@ -15,7 +15,8 @@ function IO(fn) {
 IO.prototype.chain = function(f) {
   var io = this;
   return new IO(function() {
-    return f(io.fn()).fn();
+    var next = f(io.fn.apply(io, arguments));
+    return next.fn.apply(next, arguments);
   });
 };
 
@@ -45,13 +46,6 @@ IO.prototype.of = function(x) {
 };
 
 IO.of = IO.prototype.of;
-
-// this is really only to accommodate testing ....
-IO.prototype.equals = function(that) {
-  return this === that ||
-    this.fn === that.fn ||
-    R.equals(IO.runIO(this), IO.runIO(that));
-};
 
 IO.prototype.toString = function() {
   return 'IO(' + R.toString(this.fn) + ')';
--- a/src/Reader.js
+++ b/src/Reader.js
@@ -40,12 +40,6 @@ Reader.of = Reader.prototype.of;
 
 Reader.ask = Reader(R.always);
 
-Reader.prototype.equals = function(that) {
-  return this === that ||
-  this.run === that.run ||
-  R.equals(Reader.run(this), Reader.run(that));
-};
-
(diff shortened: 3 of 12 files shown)


buzzdecafe:
LGTM. `IO.equals` always felt pretty weird

TheLudd:
If it is possible I would like to set ramda version to `>=0.15.0` so that we don't force any specific version.

davidchambers:
> If it is possible I would like to set ramda version to `>=0.15.0` so that we don't force any specific version.

How would we handle incompatibilities between the versions? I'd hate to see this sort of thing:

```javascript
var objOf = R.objOf || R.createMapEntry;
```

TheLudd:
@davidchambers Is that used in ramda fantasy? If no, then there is no problem and ramda-fantasy can in fact use any version in the range.

davidchambers:
How would we guarantee that we don't introduce a regression by using a Ramda function whose behaviour changed at some point? We'd need the test suite to run against every supported version of Ramda. It's certainly possible, but I don't see the benefit. What do we stand to gain?

TheLudd:
Every time I want to use `ramda-fantasy` I must also align the ramda version in order to not bring in duplicate versions of ramda. Thinking mostly of usage in the browser here.

Until now I could not upgrade to ramda > `0.15.x` because of this. Sure we could always do an immediate upgrade here as soon as there is a new ramda version. Perhaps that is an easier solution. But it would also force a ramda update in the project using the two modules.

You are right in that ensuring comparability against all versions would be a bit of a pain, but I could be willing to see how it works. What do y'all think would be the best solution?

joneshf:
We could just not depend on ramda at all. This is just a dependency problem, and like all dependency problems a more stable thing (this library) should not depend on a more volatile thing (ramda). There are plenty of other ways to get the same behavior until ramda settles down.

Barring that, semver ranges are good things to support. If the things used in the library don't change often enough, you can probably support a wide range. Also remember it's not just limited to `^` and `~`, you can specify a lower and upper bound. So find the oldest version that works, then go up to `0.17.1` or whatever the latest is, if it still works.

TheLudd:
Oldest current suported is `0.15.0` for the record.

scott-christopher:
I've updated ramda's version range to `>=0.15.0 <0.18.0`.

TheLudd:
I'm good with this. I acknowledge the possible problems pointed out by @davidchambers but I am willing to see how this works.

scott-christopher:
If the version range becomes an issue, we can revisit it at that time.

buzzdecafe:
i'm inclined to agree with @joneshf that removing the dependency would be a win if we can manage it

TheLudd:
@buzzdecafe @joneshf Although I thought that the purpose of this lib was to provide fantasy-land compatible types that work well with ramda which would assume that the user probably imports ramda as well. For that reason I have not hesitated to bring in ramda where it would smooth things up a bit.

joneshf:
The removal was an off-hand remark. I didn't actually look to see what all was being used.

Looking at it, I think the things that couldn't easily be removed would be the `curry`/`curryN` functions, the `equals`, and the `Future` stuff. So it's probably not worth it.

PR #67, opened 2015-08-15
Adds sum and product types.

scott-christopher:
This changes adds support for creating sum and product ADTs. I'm not entirely sure this belongs in ramda-fantasy or not, but it does make declaring data types clean and consistent (at least in my opinion).

I've also got a separate branch with the existing types converted to using ADTs for those who are curious: https://github.com/scott-christopher/ramda-fantasy/blob/adt/src/Maybe.js (I wanted to keep this PR focussed on the ADT implementation itself, before discussing changing anything else).

I've added the comments from the code below that describes the general usage.

---
## Product types

Product types can be created by calling `product` with strings representing the fields to exist on each type.

For example, the following will create a `Point` type with two fields (`x` and `y`):

``` js
var Point = Type.product('x', 'y');
```

Instances are created by calling the type as a function:

``` js
var origin = Point(0, 0);
origin instanceof Point // true
```

Once an instance is created, its fields can be accessed via getter methods attached to the instance:

``` js
var originX = origin.x();
```

Alternatively an instance can be unapplied, providing access to all its fields in one call:

``` js
origin.unapply(function(x, y) {
  return x === 0 && y === 0;
});
```

Values of a product instance can also be modified, however because the instances are immutable this will cause a new instance to be returned with the modified value rather than mutating the previous instance.

``` js
var x1y0 = origin.x(1);
x1y0.x(); // 1

// note `origin` remains unmodified
origin.x(); // 0
```

When a type is created, lenses for each field are also available on the type constructor:

``` js
R.view(Point.x, point); // 0
R.set(Point.y, point);  // Point(0, 1);
```
## Sum Types

Sum types can be created by calling `sum` with an object that describes its constructors.

For example will create a `List` type, with two constructors `Cons` and `Nil`:

``` js
var List = Type.sum({ Cons: ['head', 'tail'], Nil: [] });
var Cons = List.Cons;
var Nil = List.Nil;
```

The constructors behave exactly the same as product type constructors:

``` js
var abc = Cons('a', Cons('b', Cons('c', Nil)));
abc instanceof Cons; // true
abc instanceof List; // true

var c = abc.tail().tail().head();
// or
var c = R.view(R.compose(Cons.tail, Cons.tail, Cons.head), abc);

abc.unapply(function(head, tail) {
  return head + head + head;
}); // "aaa"
```

If a constructor is declared with no fields, a singleton instance of that constructor is returned instead:

``` js
Nil instanceof List; // true
Nil.unapply(function() { return '?'; }); // '?'
```

An instance can also be matched and dispatched against its type:

``` js
function headOrElse(other, list) {
  return list.match({
    Cons: function(head, tail) { return head; },
    Nil: function() { return other; }
  });
}
```

### DIFF
--- a/src/Type.js
+++ b/src/Type.js
@@ -1,0 +1,193 @@
+var R = require('ramda');
+
+module.exports = {
+  sum: sum,
+  product: productOf(null)
+};
+
+/**
+ * # Algebraic Data Types
+ *
+ * Algebraic data types are a way of modeling data, typically in two forms known
+ * as product types and sum types. Product types associate values with their
+ * fields (similar to a standard object in javascript), while sum types model a
+ * group of different types together as a single type.
+ *
+ * ## Product types
+ *
+ * Product types can be created by calling `product` with strings
+ * representing the fields to exist on each type.
+ *
+ * For example, the following will create a `Point` type with two
+ * fields (`x` and `y`):
+ * ```js
+ * var Point = Type.product('x', 'y');
+ * ```
+ *
+ * Instances are created by calling the type as a function:
+ * ```js
+ * var origin = Point(0, 0);
+ * origin instanceof Point // true
+ * ```
+ *
+ * Once an instance is created, its fields can be accessed via
+ * getter methods attached to the instance:
+ * ```js
+ * var originX = origin.x();
+ * ```
+ *
+ * Alternatively an instance can be unapplied, providing access to
+ * all its fields in one call:
+ * ```js
+ * origin.unapply(function(x, y) {
+ *   return x === 0 && y === 0;
+ * });
+ * ```
+ *
+ * Values of a product instance can also be modified, however because the
+ * instances are immutable this will cause a new instance to be returned with
+ * the modified value rather than mutating the previous instance.
+ * ```js
+ * var x1y0 = origin.x(1);
+ * x1y0.x(); // 1
+ *
+ * // note `origin` remains unmodified
+ * origin.x(); // 0
+ * ```
+ *
+ * When a type is created, lenses for each field are also available on the
+ * type constructor:
+ * ```js
+ * R.view(Point.x, point); // 0
+ * R.set(Point.y, point);  // Point(0, 1);
+ * ```
+ *
+ * ## Sum Types
+ *
+ * Sum types can be created by calling `sum` with an object that describes
+ * its constructors.
+ *
+ * For example will create a `List` type, with two constructors `Cons` and `Nil`:
+ * ```js
+ * var List = Type.sum({ Cons: ['head', 'tail'], Nil: [] });
+ * var Cons = List.Cons;
+ * var Nil = List.Nil;
+ * ```
+ *
+ * The constructors behave exactly the same as product type constructors:
+ * ```js
+ * var abc = Cons('a', Cons('b', Cons('c', Nil)));
+ * abc instanceof Cons; // true
+ * abc instanceof List; // true
+ *
+ * var c = abc.tail().tail().head();
+ * // or
+ * var c = R.view(R.compose(Cons.tail, Cons.tail, Cons.head), abc);
+ *
+ * abc.unapply(function(head, tail) {
+ *   return head + head + head;
+ * }); // "aaa"
+ * ```
+ *
+ * If a constructor is declared with no fields, a singleton instance
+ * of that constructor is returned instead:
+ * ```js
+ * Nil instanceof List; // true
+ * Nil.unapply(function() { return '?'; }); // '?'
+ * ```
+ *
+ * An instance can also be matched and dispatched against its type:
+ * ```js
+ * function headOrElse(other, list) {
+ *   return list.match({
+ *     Cons: function(head, tail) { return head; },
+ *     Nil: function() { return other; }
+ *   });
+ * }
+ * ```
+ */
+
+// Allows a given `proto` to be used as the base prototype of the product
+// instances -- this is used by `sum` to extend the generated sum type without
+// resorting to setting __proto__ or relying on Object.setPrototypeOf introduced
+// in ES6.
+function productOf(proto) {
+  // The actual function used to create a product type with the given set of field names.
+  return function() {
+    var field, i;
+    var fields = arguments;
+    var Constructor = R.curryN(fields.length, function () {
+      var vals = arguments;
+      return Object.create(Constructor.prototype, {
+        unapply: {
+          value: function unapply(f) {
+            return f.apply(this, vals);
+          }
+        }
+      });
+    });
+    Constructor.prototype = Object.create(proto);
+    Constructor.prototype.constructor = Constructor;
+    for (i = 0; i < fields.length; i++) {
+      field = fields[i];
+      Constructor[field] = productLens(i);
+      Constructor.prototype[field] = getterSetter(field);
+    }
+    return Constructor;
+  };
+}
+
+// The function used to create a sum type with the given constructor descriptor.
+function sum(constructors) {
+  var Type = function () {};
+
+  function buildProto(name) {
+    return Object.create(Type.prototype, {
+      match: {
+        value: function(cases) {
+          return this.unapply(cases[name]);
+        }
+      }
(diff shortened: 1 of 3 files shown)


TheLudd:
I also doubt that this belongs in ramda-fantasy. It is not an FL type (right?) and it is not something that is intended to work specifically with the existing types (right?). 

Can you explain what benefit this brings?

I also found 3 lint errors :trollface:

scott-christopher:
It doesn't implement any of the FL spec, as it is intended to potentially be used as a base for implementing the other types in ramda-fantasy.

The main benefits are:
- the instances are immutable
- it provides a consistent approach to defining types in ramda-fantasy
```js
var Maybe = Type.sum({ Just: ['value'], Nothing: []);
```
- it offers pattern matching against types for extracting values (one of the stronger benefits I believe):
```js
// rather than accessing values via conditional guards or inspecting values like so
if (Maybe.isJust(maybeThing)) {
  x = maybeThing.value;
} else {
  x = someOtherValue;
}
// can instead be accessed via matching against types
x = maybeThing.match({ Just: R.identity, Nothing: R.always(someOtherValue) });
```
- data constructors are curried
```js
var Tuple = Type.product('_1', '_2');
R.map(Tuple('a'), [0, 1, 2]); // [Tuple('a', 0), Tuple('a', 1), Tuple('a', 2)]
```
- instances can be unapplied to extract all values via a function
```js
var Point = Type.product('x', 'y');
point(2, 4).unapply((x, y) => y / x); // 2
```

If there are still concerns about whether it should be part of ramda-fantasy (or even just demoted to the `internal` folder and not exposed via the public `index.js`), I'm happy to move it to a different project.

Alternatively, if there are concerns regarding the naming, we could potentially rename the functions to something less formal. Perhaps even just offer a singe `Type` function that accepts either an array for creating product types and an object for sum types. :shrug:

---
And the linting errors were the ones I resolved in the previous PR ;) – I have since rebased.

TheLudd:
Ok I understand the benefits and will wait for input from others. Just wanna note that this makes ramda-fantasy incompatible with ramda 0.15 ;)

src/Type.js
```diff
@@ -0,0 +1,193 @@
...
+  };
+}
+
+// The function used to create a sum type with the given constructor descriptor.
+function sum(constructors) {
+  var Type = function () {};
+
+  function buildProto(name) {
+    return Object.create(Type.prototype, {
+      match: {
```
TheLudd: Could match be named `cata` instead?

davidchambers:
This is exciting, @scott-christopher! The jQuery-like getter/setter methods don't sit well with me. Providing lenses is a neat idea! If one expects users of this library to use Ramda as well, perhaps lenses alone are sufficient.

This work has significant overlap with [union-type][1]. I'm interested to read @paldepind's thoughts on this pull request. I wonder whether we could add that library as a dependency, or whether the goal is to limit this project's dependencies to Ramda itself.


[1]: https://github.com/paldepind/union-type

paldepind:
Wow. Yes. This looks cool and a lot like union-type. I quite like how you can add methods to a type by adding the type the value constructors `prototype`. That is neat (but I'm not sure it's the best way).

Union-type has a few features that I don't see in this proposal. Like type checking for fields, support for ECMAScript 6 destructuring and a design that makes `unapply` uneeded

There are currently [some changes](https://github.com/paldepind/union-type/issues/18) on the way to union-type that will also make it possible to define fantasy land compatible types (among other things). The linked GitHub issue is quite lengthy so here is an example using the next version of union-type:

```js
function maybeMap(fn) {
  var that = this;
  return this.case({
    Nothing: this.Nothing,
    Just: function(v) { return that.Just(fn(v)); },
  }, this);
}
var Maybe = Type({Just: [T], Nothing: []}, {map: maybeMap})
var just1 = Maybe.Just(1);
var just4 = just1.map(add(3));

// Nested destructuring
const T = () => true
let {Person} = Type({Person: {parent: T, name: String, id: Number}})
let father = Person(undefined, 'Knud', 12)
let createChild = Person(father)
let me = createChild('Simon', 13)
let {parent: {name: fatherName}} = me
console.log(fatherName) //=> 'Knud'
```

One thing that I don't like about neither the example above or this PR is how `map` on `Maybe` is defined. I would like to get rid of the pattern matching. Erik Meijer once said something like "if you use pattern matching you are stupid" in a talk (a bit tongue in cheek of course but it applies in some situation). We should not have to define a `map` function that pattern matches to handle both a `Just` and a `Nothing`. It should be possible to define a `map` for `Nothing` and a `map` for `Nothing`. Just like both ramda-fantasy land Sanctuary currently does it. It is not hard to implement but I can't figure out what it should look like. Maybe something like this (inspired by this PR):

```js
var Maybe = Type({Just: [T], Nothing: []})
Maybe.Just.prototype.map = // map on Just
Maybe.Nothing.prototype.map = // map on Nothing
```

But I am not happy how that makes it impossible to verify that the values have the necessary methods upon defining the type.

@alexeygolev has been sharing some splendid ideas on these matters.

scott-christopher:
> The jQuery-like getter/setter methods don't sit well with me. Providing lenses is a neat idea! If one expects users of this library to use Ramda as well, perhaps lenses alone are sufficient.

Yeah, I'm not particularly satisfied with the overloaded getter/methods, however I wanted to provided something a little less involved than requiring the use of lenses for direct field access without having to expose the values directly on the object instances. I'd welcome suggestions for alternative options here.

> This work has significant overlap with union-type.

Indeed it does. I also took inspiration from [daggy](https://github.com/puffnfresh/daggy). The main difference being that I wanted to include immutability and curried constructors.

> I wonder whether we could add that library as a dependency, or whether the goal is to limit this project's dependencies to Ramda itself.

I'm certainly open to the idea of including something like `union-type`. While it would be nice to keep the dependencies to a minimum, I don't think we should rule anything out if it will end up providing a net benefit.

> Union-type has a few features that I don't see in this proposal. Like type checking for fields ...

I didn't particularly feel the need to include type checking in this PR. I know runtime type checking is pretty polarising around here, so I'm willing to go with the majority if there is interest to include this though the overlap between this and union-type would then increase even further.

> ... support for ECMAScript 6 destructuring and a design that makes unapply uneeded

The difference between offering destructuring directly against the object instances vs `unapply` is largely a difference in offering immutability or not. If destructuring is desirable, a `#toArray` method could be added such that one could `let [value] = just5.toArray()` or a bit more verbosely as is via `just5.unapply(R.unapply(R.identity))` for a double-dose of `unapply` :D (I still wonder whether an `R.args` function would be a useful alias in ramda for `R.unapply(R.identity))`.

> I would like to get rid of the pattern matching. Erik Meijer once said something like "if you use pattern matching you are stupid" in a talk

I believe Erik's comment (if I'm thinking of the same comment) was regarding preference of folds over the use of pattern matching and recursion. I personally prefer having the definitions for functions grouped together via the pattern matched blocks, rather than on separate prototypes, but that preference may be largely subjective. I'd also like to move to a style where the functions are defined statically on the types where possible rather than prototype methods, something that pattern matching helps with – again, this is largely subjective and I'm happy to be convinced otherwise.

src/Type.js
scott-christopher: I'm not entirely familiar with the nuances of catamorphisms, but this method is named `cata` in other similar implementations and it does make sense in terms of folding the `Type` down to some other value. Perhaps @joneshf might be able to provide some insight?

I did however use the opportunity to follow through [this intro to F-Algebras and catamorphisms](https://www.fpcomplete.com/user/bartosz/understanding-algebras), translating the examples to JS using `Type` from this PR: https://gist.github.com/scott-christopher/35568367b82db7ef22a1 :smile:

jethrolarson:
I'm interested in seeing this. /bump

buzzdecafe:
let's merge this; it is certainly interesting enough

src/Type.js
joneshf: Sorry, no insight here.

buzzdecafe:
d'oh, waited (way) too long and now this has conflicts. are you interested in resolving these @scott-christopher ?

Issue #74, opened 2015-09-27
Reader ask


davidchase:
Should `Reader.ask` equal `Reader(R.identity)` ?  so the following would work
```js
const append = str => Reader.ask.map((strToo) => str.concat(' ', strToo));
const greet = R.compose(R.chain(append), Reader.of);
greet('hello').run('world'); //=> hello world
```

rather current with `Reader.ask = Reader(R.always)`

```js
const append = str => Reader.ask.map((strToo) => str.concat(' ', strToo()));
const greet = R.compose(R.chain(append), Reader.of);
greet('hello').run('world');  //=> hello world
```

you need to call `strToo()`...

scott-christopher:
You're correct. It should be using `identity`. I'll raise a PR shortly.

Thanks for raising it.

scott-christopher:
I've raised https://github.com/ramda/ramda-fantasy/pull/76 to fix this.

Thanks again.

davidchase:
cool :+1: thanks for looking into it

TheLudd:
We could release this as a patch version

scott-christopher:
I've just pushed out 0.4.1

PR #76, opened 2015-09-29
Fixes Reader.ask

scott-christopher:
Addresses #74


TheLudd:
This is basically a revert to the way it was before but nice with the added test.

PR #77, opened 2015-10-14
Various changes to Either

### LINKED ISSUE #75: Maybe and Either have different "commute" semantics
Commuting a list of Maybes results in a Nothing if any of the elements if a Nothing:

```javascript
> R.commute(Maybe.of, [Maybe.Just(1), Maybe.Just(2)])
Maybe.Just([1, 2])
> R.commute(Maybe.of, [Maybe.Just(1), Maybe.Just(2), Maybe.Nothing()])
Maybe.Nothing()
```

Commuting a list of Eithers results in a Right whose value is a list containing each Right's value:

```javascript
> R.commute(Either.of, [Either.Right(1), Either.Right(2)])
Either.Right([1, 2])
> R.commute(Either.of, [Either.Right(1), Either.Right(2), Either.Left('XXX')])
Either.Right([1, 2])
```

scott-christopher:
- Either::ap now returns `this` for `Left` (fixes #75)
- ~~Either::extract added~~
- ~~Either::extend no longer returns `this` for `Left` to satisfy `extract . extend f = f`~~
- Either tests refactored to use JSVerify
- Comonad laws added to test/types.js

### DIFF
--- a/test/either.test.js
+++ b/test/either.test.js
@@ -1,105 +1,91 @@
+var R = require('ramda');
 var assert = require('assert');
-var equalsInvoker = require('./utils').equalsInvoker;
-var types = require('./types')(equalsInvoker);
+var jsv = require('jsverify');
+var types = require('./types')(R.equals);
 
 var Either = require('..').Either;
 
+var leftArb = function(arb) {
+  return arb.smap(Either.Left, R.prop('value'), R.toString);
+};
+var rightArb = function(arb) {
+  return arb.smap(Either.Right, R.prop('value'), R.toString);
+};
+var eitherArb = function(arb) {
+  return jsv.oneof([leftArb(arb), rightArb(arb)]);
+};
+
+var eNatArb = eitherArb(jsv.nat);
+var eFnArb = eitherArb(jsv.fn(jsv.nat));
+var fnNatArb = jsv.fn(jsv.nat);
+
 describe('Either', function() {
-  var e = Either('original left', 1);
-
-  function mult(a) {
-    return function(b) { return a * b; };
-  }
-
-  function add(a) {
-    return function(b) { return a + b; };
-  }
 
   it('is a Functor', function() {
     var fTest = types.functor;
-    assert.equal(true, fTest.iface(e));
-    assert.equal(true, fTest.id(e));
-    assert.equal(true, fTest.compose(e, mult(2), add(3)));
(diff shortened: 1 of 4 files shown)


test/either.test.js
```diff
@@ -1,86 +1,78 @@
...
-var types = require('./types')(equalsInvoker);
+var jsv = require('jsverify');
+var types = require('./types')(R.equals);
 
 var Either = require('..').Either;
 
-describe('Either', function() {
-  var e = Either('original left', 1);
+var leftArb = function(arb) {
+  return arb.smap(Either.Left, R.prop('value'), R.invoker(0, 'toString'));
```
davidchambers: Could we use `R.toString` in place of `R.invoker(0, 'toString')`?

scott-christopher: Sure.

davidchambers:
LGTM!

scott-christopher:
I would be interested to get some suggestions around the names of the comonad laws. Perhaps `id` and `idF` would be better labelled as `leftIdentity` and `rightIdentity`?

ping @joneshf

TheLudd:
Was it @joneshf that said that this version of extract that returns either the left or the right value would break comonad laws?

TheLudd:
It was!
https://gitter.im/fantasyland/fantasy-land?at=55caf75daac97ada66dd990f

scott-christopher:
I am going to plead ignorance when it comes to the comonad/extend instance for `Either` :smile:

Switching the implementation to the one included in this PR seems to satisfy the comonad laws (also included in this PR), though I'm not aware how this could be useful (perhaps an interpreter for some recursive type? :shrugs:)

After further investigation, I've found `Either` implementing `Extend` extend elsewhere but not `Comonad` (i.e. no `extract`), reflecting the existing implementation before this PR.

I'll roll back the implementation in this PR to only include `extend`.

scott-christopher:
I've amended the PR to remove `Either::extract` and revert the implementation of `Left::extend`.

joneshf:
> I would be interested to get some suggestions around the names of the comonad laws. Perhaps id and idF would be better labelled as leftIdentity and rightIdentity?

Sure. I think the `Extend` law is actually associativity though.

In any case, this is great!

scott-christopher:
Thanks @joneshf, I've changed the extend law to `associative`.

Any other concerns with this change? It'd be good to get the `ap` fix in.

PR #79, opened 2015-10-24
Improve Future try-catching

Avaq:
I noticed Ramda Future wraps `fork` in a *try-catch*. I'm assuming this was done with the intention to automatically catch errors from throwing user functions. In a purely functional world that wouldn't make sense, but I can appreciate the idea that this protects users of Future from the rigors of multi-paradigm.

However, I think that Futures should either *have* this behaviour (like `bluebird`), or not have it at all (like `data.task`), but Ramda Futures have it "*sometimes*", as demonstrated by the following code (also added as unit test), and I think this is quite confusing:

```js
//The error is caught and handled by handler:
Future.of(1).map(throws).fork(handler, noop);

//The error escapes and destroys everything you ever loved:
delayedFutureOf(1).map(throws).fork(handler, noop);
```
> Open debugger to see effects: http://goo.gl/crHJ5F

Another issue with the current implementation is that often, functions that we know are *pure* are also being wrapped in try-catch statements. This [probably](https://github.com/ramda/ramda-fantasy/pull/82) results in unnecessary performance loss.

I tried to solve both issues with the following changes:

* ~~Switched many internal "fork" calls to "_fork" calls to reduce needless try-catch wrapping of internal functions that were never going to throw.~~ (see [pull 82](https://github.com/ramda/ramda-fantasy/pull/82))
* Wrapped potentially throwing user functions in try-catch to ensure the error will end up in the rejection branch.

Another solution would be to never try-catch, which could be implemented by simply removing `Future#fork` and assigning it in the constructor.

> #### Disclaimer
> I haven't actually tested for performance gains or losses. :)

### DIFF
--- a/src/Future.js
+++ b/src/Future.js
@@ -1,5 +1,15 @@
 var R = require('ramda');
 
+function jail(handler, f){
+  return function(a){
+    try{
+      return f(a);
+    } catch(err) {
+      handler(err);
+    }
+  };
+}
+
 // `f` is a function that takes two function arguments: `reject` (failure) and `resolve` (success)
 function Future(f) {
   if (!(this instanceof Future)) {
@@ -9,11 +19,7 @@ function Future(f) {
 }
 
 Future.prototype.fork = function(reject, resolve) {
-  try {
-    this._fork(reject, resolve);
-  } catch(e) {
-    reject(e);
-  }
+  this._fork(reject, jail(reject, resolve));
 };
 
 // functor
@@ -29,11 +35,11 @@ Future.prototype.ap = function(m) {
     var applyFn, val;
     var doReject = R.once(rej);
 
-    function resolveIfDone() {
+    var resolveIfDone = jail(doReject, function() {
       if (applyFn != null && val != null) {
         return res(applyFn(val));
       }
-    }
+    });
 
     self.fork(doReject, function(fn) {
       applyFn = fn;
@@ -64,8 +70,10 @@ Future.prototype.of = Future.of;
 //:: Future a, b => (b -> Future c) -> Future c
 Future.prototype.chain = function(f) {  // Sorella's:
   return new Future(function(reject, resolve) {
-    return this.fork(function(a) { return reject(a); },
-                     function(b) { return f(b).fork(reject, resolve); });
+    return this.fork(
+      function(a) { return reject(a); },
+      jail(reject, function(b) { return f(b).fork(reject, resolve); })
+    );
   }.bind(this));
 };
 
@@ -74,9 +82,10 @@ Future.prototype.chain = function(f) {  // Sorella's:
 //:: Future a, b => (a -> Future c) -> Future c
 Future.prototype.chainReject = function(f) {
   return new Future(function(reject, resolve) {
-    return this.fork(function(a) { return f(a).fork(reject, resolve); },
-                     function(b) { return resolve(b);
-    });
+    return this.fork(
+      jail(reject, function(a) { return f(a).fork(reject, resolve); }),
+      function(b) { return resolve(b); }
+    );
   }.bind(this));
 };
 
@@ -87,11 +96,10 @@ Future.prototype.chainReject = function(f) {
 Future.prototype.bimap = function(errFn, successFn) {
   var self = this;
   return new Future(function(reject, resolve) {
-    self.fork(function(err) {
-      reject(errFn(err));
-    }, function(val) {
-      resolve(successFn(val));
-    });
+    self.fork(
+      jail(reject, function(err) { reject(errFn(err)); }),
+      jail(reject, function(val) { resolve(successFn(val)); })
+    );
   });
 };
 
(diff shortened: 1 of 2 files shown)


buzzdecafe:
sorry for the delay -- this makes sense to me. Anyone else have input? @TheLudd ? @scott-christopher ?

Avaq:
Perhaps I should PR a version where I only did the first of the two bullet-points (the performance optimization). Because I can imagine the second doesn't sit well with people. We don't want to encourage using `map(mightThrow)` over `chain(mightReject)` or `map(returnsMaybe)`. It doesn't take away my wonder why the *try-catch* is there in the first place though.

TheLudd:
Sorry for being inactive on this. 

The reason for it being added was that when errors occurred in mapping functions they completely disappeared and were hard to trace. I understand it might not be the fully functional way. I am certainly up for discussions on how to handle it differently. 

But you found a bug and it should certainly be fixed. I would however prefer the bug fix and the optimization to be merged/discussed separately so that discussions can be more focused and possible revers easier.

Avaq:
The reason the bug fix and the optimization are in the same commit is because they were part of the same problem in my head: "The wrong functions are being try-catch-wrapped in the wrong places". I could tear the optimization out to make this PR just about the bug fix and try-catch discussion.

There's not much to discuss about the performance optimization. It's just calling `_fork` instead of `fork` in places where we are in control of the full `res` stack so we can be sure it wouldn't have thrown anyway.

Avaq:
Problem is: I can only be sure "it wouldn't have thrown anyway", if I can be assured that nested `res` calls won't throw, which I only can if this bug is fixed.

Avaq:
Alright. I've separated the performance thing into [its own PR](https://github.com/ramda/ramda-fantasy/pull/82) :)

Avaq:
> I am certainly up for discussions on how to handle it differently.

My opinion is as stated in the PR description: I think you should do it *well* or not do it at all.

The do-it-well case has the advantage that it protects users from these difficult to track errors that even have the potential to crash your app because nobody thought to catch them. The disadvantages are:

- We are committing to try-catch wrapping *every function stack* that we are executing *after every potential JavaScript tick* (so every time a callback could have gone async). This means we're try-catch wrapping way more than what would be necessary for a sane application "just to be safe".
- We might encourage writing code like: `myFuture.map(throwIfInvalid).map(valid => ...)` and it would behave differently from any other Functor.

The not-at-all case (which `data.task` has opted for) has the advantage that the user is put in control over which function they should try-catch-wrap, which means they can be much more selective in where to apply these wraps. The disadvantage is that the user will have a much harder time to 1. Make safe code and 2. Deal with failure when it does happen.

We may also want to offer this wrapping as a ulility, if we were going to choose the not-at-all case, to minimize the amount of work a user has to do to make their code throw-safe. Instead of `m.map(mightThrow)` a user would do something like `m.chain(Future.jail(mightThrow))`.

buzzdecafe:
> We may also want to offer this wrapping as a ulility, if we were going to choose the not-at-all case, to minimize the amount of work a user has to do to make their code throw-safe. Instead of `m.map(mightThrow)` a user would do something like `m.chain(Future.jail(mightThrow))`.

Of the three options you presented, this one is my preference; it gets the benefits of "no-at-all" while providing a way for the user to clean up his own mess.

Avaq:
It is worth noting that the not-at-all case breaks backwards compatibility, while the do-it-well case does not.

buzzdecafe:
> It is worth noting that the not-at-all case breaks backwards compatibility, while the do-it-well case does not.

i'd rather get it right now than be beholden to prior design decisions

Avaq:
Shall I make Yet Another PR® to offer `jail` as a utility? The PR would obsolete #79 and #82.

Might be worth having a quick talk about the name of the function too. Sanctuary calls it `encase` rather than `jail`. We could go with `Future.encase` or `encaseFuture` or `Future.jail`. Maybe it's not even Ramda-Fantasy's responsibility to provide this utility, but it does also provide `Future.cache`.

TheLudd:
I'd rather stay on course and do it the way we do now and do it well.
There has been discussions about handling Futures in other ways, such as creating a `FutureTEither` type that does not do any forking but always resolves to an `Either`. If this is done perhaps there will be better or more pure ways to handle errors. 

The current solution might not be perfect but until we know what we *really* want I would not encourage breaking changes.

buzzdecafe:
> The current solution might not be perfect but until we know what we really want I would not encourage breaking changes.

i'll defer to @TheLudd on this.

TheLudd:
Ok. This PR looks good to me. Perhaps a squash?

@scott-christopher Last chance to comment....

scott-christopher:
Thanks @Avaq, this LGTM if you can squash the last commit.

Avaq:
@TheLudd @scott-christopher I've squashed my two commits into one. I believe it should be good to merge. :)

scott-christopher:
Thanks @Avaq

PR #81, opened 2015-11-18
specify "@@type" property for each type

davidchambers:
This provides compatibility with `S.type`, added in plaid/sanctuary#105:

> #### `type :: a -> String`
>
> Takes a value, `x`, of any type and returns its type identifier. If `x` has a `'@@type'` property whose value is a string, `x['@@type']` is the type identifier. Otherwise, the type identifier is the result of applying [`R.type`][1] to `x`.
>
> `'@@type'` properties should use the form `'<package-name>/<type-name>'`, where `<package-name>` is the name of the npm package in which the type is defined.
>
> ```javascript
> > S.type(S.Just(42))
> 'sanctuary/Maybe'
>
> > S.type([1, 2, 3])
> 'Array'
> ```

[1]: http://ramdajs.com/docs/#type

### DIFF
--- a/src/Either.js
+++ b/src/Either.js
@@ -16,6 +16,8 @@ function Either(left, right) {
   }
 }
 
+Either.prototype['@@type'] = 'ramda-fantasy/Either';
+
 Either.prototype.map = util.returnThis;
 
 Either.of = Either.prototype.of = function(value) {
--- a/src/Future.js
+++ b/src/Future.js
@@ -18,6 +18,8 @@ function Future(f) {
   this._fork = f;
 }
 
+Future.prototype['@@type'] = 'ramda-fantasy/Future';
+
 Future.prototype.fork = function(reject, resolve) {
   this._fork(reject, jail(reject, resolve));
 };
--- a/src/IO.js
+++ b/src/IO.js
@@ -11,6 +11,8 @@ function IO(fn) {
   this.fn = fn;
 }
 
+IO.prototype['@@type'] = 'ramda-fantasy/IO';
+
 // `f` must return an IO
 IO.prototype.chain = function(f) {
   var io = this;
--- a/src/Identity.js
+++ b/src/Identity.js
@@ -20,6 +20,8 @@ function Identity(x) {
   this.value = x;
 }
 
+Identity.prototype['@@type'] = 'ramda-fantasy/Identity';
+
 /**
  * Applicative specification. Creates a new `Identity[a]` holding the value `a`.
  * @param {*} a Value of any type
--- a/src/Maybe.js
+++ b/src/Maybe.js
@@ -6,6 +6,8 @@ function Maybe(x) {
   return x == null ? _nothing : Maybe.Just(x);
 }
 
+Maybe.prototype['@@type'] = 'ramda-fantasy/Maybe';
+
 function Just(x) {
   this.value = x;
 }
--- a/src/Reader.js
+++ b/src/Reader.js
@@ -12,6 +12,8 @@ Reader.run = function(reader) {
   return reader.run.apply(reader, [].slice.call(arguments, 1));
 };
 
+Reader.prototype['@@type'] = 'ramda-fantasy/Reader';
+
 Reader.prototype.chain = function(f) {
   var reader = this;
   return new Reader(function(r) {
--- a/src/Tuple.js
+++ b/src/Tuple.js
@@ -36,6 +36,8 @@ Tuple.snd = function(x) {
   return x[1];
 };
 
+_Tuple.prototype['@@type'] = 'ramda-fantasy/Tuple';
+
 // semigroup
 _Tuple.prototype.concat = function(x) {
   ensureConcat([this[0], this[1]]);
--- a/test/either.test.js
+++ b/test/either.test.js
@@ -110,6 +110,15 @@ describe('Either', function() {
     });
   });
 
+  describe('#@@type', function() {
+
+    it('is "ramda-fantasy/Either"', function() {
+      assert.strictEqual(Either.Left(null)['@@type'], 'ramda-fantasy/Either');
+      assert.strictEqual(Either.Right(null)['@@type'], 'ramda-fantasy/Either');
+    });
+
+  });
+
   describe('#toString', function() {
 
     it('returns the string representation of a Left', function() {
--- a/test/future.test.js
+++ b/test/future.test.js
@@ -191,6 +191,17 @@ describe('Future', function() {
 
   });
 
+  describe('#@@type', function() {
+
+    it('is "ramda-fantasy/Future"', function() {
+      assert.strictEqual(
+        Future(function(reject, resolve) { void resolve; })['@@type'],
+        'ramda-fantasy/Future'
+      );
+    });
+
+  });
+
   describe('#toString', function() {
 
     it('returns the string representation of a Future', function() {
--- a/test/identity.test.js
+++ b/test/identity.test.js
@@ -62,6 +62,14 @@ describe('Identity', function() {
     assert.equal(true, mTest.iface(m));
   });
 
+  describe('#@@type', function() {
+
+    it('is "ramda-fantasy/Identity"', function() {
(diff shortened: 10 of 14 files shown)


CrossEye:
Trying to catch up on things.  This is another one I missed.

I'm wondering if this should use `Symbol.for('type')` in environments that support it before `@@type`.  I suppose that's a question for Sanctuary, not Ramda-Fantasy, though, so I'll go raise it there, if it's not already part of the discussion.

buzzdecafe:
`Symbol.for('type')` seems to me to squat on a pretty common term. 

The larger issue I have is that glomming this property onto the prototype feels like Flag Oriented Programming, with the "magic string" extension. Maybe this is the only way to get JS to behave in a typed fashion -- or maybe we're pushing it beyond the limits of its ability.

CrossEye:
> The larger issue I have is that glomming this property onto the prototype feels like Flag Oriented Programming, with the "magic string" extension. 

I'm afraid that might end up being the case, but I'd like to see how it goes.  It seems an interesting experiment at least.

davidchambers:
It's nice to be able to determine that `Just(42)` and `Nothing()` are members of the same type, because they have the same value for a specially named property. This is preferable to them sharing an "abstract base class" for several reasons:

  - introducing the concept of subtypes is confusing when what we *actually* want are union types with multiple constructors;

  - `instanceof` checks are brittle; and

  - mandating `instanceof` support places unnecessary restrictions on ADT implementations.

scott-christopher:
Is the primary goal of this to provide a more reliable way of determining whether an instance is a member of a given sum type, or to use as the keys for dispatching (or perhaps both)?

buzzdecafe:
> This is preferable to them sharing an "abstract base class"

not arguing that. i don't think there is a "good" solution

davidchambers:
> Is the primary goal of this to provide a more reliable way of determining whether an instance is a member of a given sum type, or to use as the keys for dispatching (or perhaps both)?

The former. The latter had not occurred to me, but it's an interesting idea!

scott-christopher:
Perhaps another option would be to use a static method on the types or standalone function to determine whether an instance belongs to the type.

e.g.
```js
// isTypeOf :: * -> Boolean
Maybe.isTypeOf(Just(42)); // true
Maybe.isTypeOf(Left(24)); // false

// or namespaced

Maybe['@@type/is'](Just(42)); // true
```
This could potentially be used by `R.is` too.

I'd also prefer to have fewer reasons for needing this check to begin with, though I appreciate it has its place for things such as asserting contracts at runtime.

davidchambers:
That would work, @scott-christopher, though I would miss being able to ask _What **are** you?_ in a REPL session rather than asking _Are you a Foo?_, _Are you a Bar?_, etc. A suitable `toString` method helps to answer this question, of course, but only if one recognizes the names of a type's data constructors.

buzzdecafe:
> though I would miss being able to ask _What are you?_

you could have a generic utility function type -> instance -> bool, and then the static, type-specific version would simply curry in the type.

davidchambers:
@buzzdecafe, that seems to answer _Are you a &lt;blank&gt;?_ rather than _What **are** you?_ I want to be able to ask `S.type(x)` in the REPL and be told `'Array'` or `'sanctuary/Maybe'` or `'ramda-fantasy/IO'`. :)

mattferrin:
I'm an outsider, but it appears to me that @scott-christopher and @buzzdecafe are correct to what to ask the question "are you a <blank>" because doing so is more general and flexible.

If you do choose to ask "what are you", I would suggest doing so by passing the instance or the type into a function that answers the question by returning a list of types the instance or type satisfies. You could do so by folding over a list of functions that return the answer to "are you a <blank>" (basically).

CrossEye:
My problem is that I don't want to ask the question "what are you?", but neither do I want to ask, "Are you a `flubnitz/Quizdoodle`?" / "Are you a `flubnitz/Krunforz`?"

What I really want to know is "Are you a `Monoid`?" / "Are you a `Functor`?"  And then I might have a few follow-up questions...

mattferrin:
@CrossEye That seems like a great case for "flag oriented" programming at first glance, especially since it would allow you test all relevant types to insure they follow the respective rules they claim to implement. Great point.

buzzdecafe:
so w/r/t this PR, where do we stand? is `@@type` better than what we currently have (i.e. not much -- `isJust` et al)?

CrossEye:
>  is `@@type` better than what we currently have (i.e. not much -- `isJust` et al)?

I'm afraid it might be.  But it feels like the wrong direction.

davidchambers:
`@@type` does not compete with functions/properties such as `isNothing` and `isJust`. `@@type` allows us to determine whether we've been given a value of the expected type; `isNothing` and `isJust` allow us to make decisions based on a known (or assumed) good value.

CrossEye:
>> Is the primary goal of this to provide a more reliable way of determining whether an instance is a member of a given sum type, or to use as the keys for dispatching (or perhaps both)?

> The former. The latter had not occurred to me, but it's an interesting idea!

That's interesting, because I remember this coming up in the context of @jethrolarson's work on a dispatching mechanism for something very closely related to the FantasyLand specification.  As I was working on a similar system myself at the time, I was very interested in his approach.  At some point when several other side projects wind down, I'd like to get back to that.

scott-christopher:
To enumerate the use cases listed so far along with my thoughts/ramblings of each:

1. **Determine whether something belongs to some type**
   * As already mentioned, I'd probably prefer a static method to take care of this, as I'd be concerned to start seeing stringly-typed code such as `if ($thing['@@type'] === 'FlizzleBlub') ...`

   * However a static method would potentially be cursed with the same `instanceOf` concerns to only being able to discern type membership for instances belonging to the same versions of a given type. e.g. `RFv0.4_Maybe['@@type/is'](RFv0.4.1_Just(x)); // potentially false` – one solution to this problem is to compare strings (such as the change to `R.__`).

   * Ideally I'd prefer the question of `are you a ...?` to not exist in the first place
2. **Retrieve a string representation of the type for a given instance for debugging**

   * This would be useful, though I wonder whether the type information could also be captured in `toString`
3. **Ask whether something implements a given typeclass / dispatching in polymorphic functions**
   * I think I'd prefer to have namespaced methods such as `@@fl/chain` or `@@data/monad/chain`, etc to indicate supported typeclasses. I can't see a reason why we can't lead this among Ramda/RF/Sanctuary, even if it just aliases back to the existing methods for now.

   * That said, I believe there are a few people currently looking into ways to decouple the typeclass implementations from the prototype of the datatype. Using an `@@type` property would be one way to achieve that, though again, I think I'd prefer a composition of predicate functions to determine dispatching. Comparing the string value of a property means there wouldn't be a nice way to then define implementations for built-in types without modifying their prototypes, while using predicates an environment would effectively just become a series of `R.ifElse` for each registered implementation, throwing a `TypeError` if nothing returned true.

CrossEye:
@scott-christopher:

This seems very relevant to your last point: https://github.com/fantasyland/fantasy-land/issues/92

> I think I'd prefer a composition of predicate functions to determine dispatching.

That was the main difference between my approach and @jethrolarson's.

buzzdecafe:
thanks for refreshing my memory about that thread @CrossEye . i also agree that the more interesting questions to ask are _What type class are you?_ which might be made easier(?) with namespaced FL methods. then `areYouAFunctor f -> is(Function, f["@@fantasyland-type/map"])` or some such

davidchambers:
Let's merge or close this pull request. :)

buzzdecafe:
in spite of my own objections -- or squeamishness, at least -- i think we should merge to facilitate merging the repos generally. Anyone object? Speak now or forever hold yer peace

ping @TheLudd @CrossEye @scott-christopher

buzzdecafe:
... and apologies for dropping the ball

scott-christopher:
I don't see any harm in merging this. If something different materialises out of FL in the future that can cover these use cases then we can adapt to it at that point in time.

PR #82, opened 2015-11-25
Reduce on unnecessary try-catch wrapping in Futures

Avaq:
Switched many internal `fork` calls to `_fork` calls to reduce needless try-catch wrapping of internal functions that were never going to throw. Taken from [pull 79](https://github.com/ramda/ramda-fantasy/pull/79) where it was originally introduced.

The reason behind this change is that *try-catch-wrapping* [might cost some performance](https://jsperf.com/try-catch/7). I haven't actually tested this for performance gains, but I'd imagine that all that try-catch wrapping when using Future-heavy code might add up.

### DIFF
--- a/src/Future.js
+++ b/src/Future.js
@@ -41,12 +41,12 @@ Future.prototype.ap = function(m) {
       }
     });
 
-    self.fork(doReject, function(fn) {
+    self._fork(doReject, function(fn) {
       applyFn = fn;
       resolveIfDone();
     });
 
-    m.fork(doReject, function(v) {
+    m._fork(doReject, function(v) {
       val = v;
       resolveIfDone();
     });
@@ -70,9 +70,9 @@ Future.prototype.of = Future.of;
 //:: Future a, b => (b -> Future c) -> Future c
 Future.prototype.chain = function(f) {  // Sorella's:
   return new Future(function(reject, resolve) {
-    return this.fork(
+    return this._fork(
       function(a) { return reject(a); },
-      jail(reject, function(b) { return f(b).fork(reject, resolve); })
+      jail(reject, function(b) { return f(b)._fork(reject, resolve); })
     );
   }.bind(this));
 };
@@ -82,8 +82,8 @@ Future.prototype.chain = function(f) {  // Sorella's:
 //:: Future a, b => (a -> Future c) -> Future c
 Future.prototype.chainReject = function(f) {
   return new Future(function(reject, resolve) {
-    return this.fork(
-      jail(reject, function(a) { return f(a).fork(reject, resolve); }),
+    return this._fork(
+      jail(reject, function(a) { return f(a)._fork(reject, resolve); }),
       function(b) { return resolve(b); }
     );
   }.bind(this));
@@ -96,7 +96,7 @@ Future.prototype.chainReject = function(f) {
 Future.prototype.bimap = function(errFn, successFn) {
   var self = this;
   return new Future(function(reject, resolve) {
-    self.fork(
+    self._fork(
       jail(reject, function(err) { reject(errFn(err)); }),
       jail(reject, function(val) { resolve(successFn(val)); })
     );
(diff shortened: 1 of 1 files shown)


src/Future.js
```diff
@@ -35,12 +35,12 @@ Future.prototype.ap = function(m) {
...
     }
 
-    self.fork(doReject, function(fn) {
+    self._fork(doReject, function(fn) {
       applyFn = fn;
       resolveIfDone();
     });
 
-    m.fork(doReject, function(v) {
+    m._fork(doReject, function(v) {
```
Avaq: The consequence of this line, I just realized, is that it killed interoperability with other potential Futures that implement the "*standard*" of a `.fork(rej, res)`. I'm not sure this is of too big a concern, but it might annoy someone somewhere.

Avaq:
I've rebased and resolved merge conflicts. @TheLudd, @scott-christopher, can I get your thoughts on the interoperability of Futures (or any 3d party fantasy types for that matter), as pointed out in the above comment? That would also be useful to keep in mind future PR's I might make. :)

Avaq:
In a way it's breaking backwards-compatibility, albeit very subtly. We're treating input in a different manner from how we used to. It might even break someones unit tests where they mock a Future. So I'm quite convinced at this point that I should undo the changes in places where we can't be reasonably sure to have a `_fork()` available.

Avaq:
Alternatively I could let the code decide at runtime by type-checking for our own Future. Might be worth the performance gains. I'm sure that the passed Future will be a Ramda Future in most if not all of the cases.

src/Future.js
```diff
@@ -10,6 +10,16 @@ function jail(handler, f){
...
 }
 
+function fork(m, reject, resolve) {
+  if(m instanceof Future) {
+    return m._fork(reject, resolve);
+  } else if (typeof m.fork === 'function') {
+    return m.fork(reject, resolve);
+  } else {
+    reject(new TypeError('Expected a Future, got ' + R.toString(m)));
+  }
```
Avaq: It would have thrown "Undefined is not a function", may as well make it a little more clear while we're at it. :)

src/Future.js
scott-christopher: I'm not too concerned about interop between other "forkables", as to the best of my knowledge that has never been a goal of `Future`. The type passed to `ap` should also be of the same type as the instance whose method is being called, so this isn't an issue for me (but thanks for raising).

src/Future.js
```diff
@@ -10,6 +10,16 @@ function jail(handler, f){
   };
 }
 
+function fork(m, reject, resolve) {
```
scott-christopher: As mentioned in my other comment, I wouldn't be concerned with trying to make this play nicely with other things that look like a `Future` for now. If we were to go down that path, we would first need to spend some time defining a spec around expected behaviour.

Avaq: I figured this'd be necessary to preserve backwards compatibility (as well as interoperability). For example it could break existing unit tests where someone mocked a Future by `{fork: sinon.spy()}` or something. I guess we could say that this "is their problem" for not passing the right type. I'll undo this last commit if you want.

scott-christopher: I think for that particular case my suggestion would be to use `Future(sinon.spy())` instead.

I'm personally in favour of sticking with just the first commit in this PR, but I welcome input from others. /cc @buzzdecafe @TheLudd

TheLudd: I agree that there is no need for type checking. If we would want to do that then we should do it everywhere in all types.

Avaq: I undid this commit.

src/Future.js
```diff
@@ -133,7 +133,7 @@ Future.cache = function(f) {
 
   function doResolve(reject, resolve) {
     status = 'PENDING';
-    return f.fork(
+    return f._fork(
```
Avaq: After some thinking and code-path tracing I realized handleCompletion does not deal with user-land functions that haven't been "jailed" already. So it should be safe to apply the optimization here too.

scott-christopher:
This looks good to me now.

@Avaq: Can you just squash the extra commit and then it's good to merge.

Avaq:
@scott-christopher :+1:

scott-christopher:
Thanks again @Avaq.

PR #83, opened 2015-12-05
Add Future#concat()

Avaq:
**Includes a failing unit test for discussion**

Hello again,

I've been working a tremendous amount with Futures of late. I've always liked Promises since their earliest introduction. Futures to me are the next step. They have a predictable API which is very easy to abstract over and therefore they integrate very well with stuff like Ramda.

However, despite being able to express all manner of async behaviour using just `chain` for sequences and `ap` for parrallelism, I find myself really missing a native, chainable way to do races. This is where `concat` comes in, which was heavily inspired by [its implementation in `data.task`](https://github.com/folktale/data.task/blob/master/lib/task.js#L137-L175).

Sorella commented it to be a Semigroup, as is also indicated by my passing *is a Semigroup* -test. Yet I've never really trusted it to always behave this way under race conditions, so I decided to write another test: *associativity is not broken by race-conditions*, which turned out to fail.

After some days of this branch collecting dust, I still decided to make a PR just to share my findings and maybe someone might yet point out that I'm wrong somehow and Sorella was right about her assessment of this implementation being a proper Semigroup. That's what I'm secretly hoping for, because it'd be really nice to get Semigroup and async races in Futures.

If, however, we collectively reach the conclusion that Future being Semigroup is in fact impossible, I'll gladly back down and implement this function as a helper in my code base somewhere.

### DIFF
--- a/test/future.test.js
+++ b/test/future.test.js
@@ -23,10 +23,44 @@ Future.prototype.equals = function(b) {
 
 describe('Future', function() {
 
+  function delayError(delay, err) {
+    return new Future(function(reject) {
+      setTimeout(reject, delay, err);
+    });
+  }
+
+  function delayValue(delay, value) {
+    return new Future(function(reject, resolve) {
+      setTimeout(resolve, delay, value);
+    });
+  }
+
+  function assertCbVal(done, expectedVal) {
+    return function(val) {
+      assert.equal(expectedVal, val);
+      done();
+    };
+  }
+
+  function fail(done){
+    return function(x){
+      done(new Error('Forked to unexpected branch with value: ' + R.toString(x)));
+    };
+  }
+
   it('should equal another future', function() {
     var f1 = Future.of(2);
     var f2 = Future.of(2);
     assert.equal(true, f1.equals(f2));
+  });
+
+  it('is a Semigroup', function() {
+    var sTest = types.semigroup;
+    var good = Future.of(2);
+    var bad = Future.reject(-2);
+    var slow = delayValue(10, 1);
+    assert.equal(true, sTest.iface(good));
+    assert.equal(true, sTest.associative(good, bad, slow));
   });
 
   it('is a Functor', function() {
@@ -98,25 +132,6 @@ describe('Future', function() {
   describe('#ap', function() {
     /*jshint browser:true */
     var add = R.add;
-    function delayError(delay, err) {
-      /*jshint unused:false */
-      return new Future(function(reject, resolve) {
-        setTimeout(reject, delay, err);
-      });
-    }
-
-    function delayValue(delay, value) {
-      return new Future(function(reject, resolve) {
-        setTimeout(resolve, delay, value);
-      });
-    }
-
-    function assertCbVal(done, expectedVal) {
-      return function(val) {
-        assert.equal(expectedVal, val);
-        done();
-      };
-    }
 
     it('applies its function to the passed in future', function() {
       var f1 = Future.of(add(1));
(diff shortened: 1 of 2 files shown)


test/future.test.js
```diff
@@ -191,6 +206,40 @@ describe('Future', function() {
...
+
+      var f1 = randomDelayValue(1);
+      var f2 = randomDelayValue(2);
+      var f3 = randomDelayValue(3);
+
+      for(i=0; i<20; i++){
+        assert.equal(true, assoc(f1, f2, f3));
+      }
+
+      setTimeout(done, 50);
```
Avaq: Because of the way `Future.prototype.equals` is implemented, and the way I need to perform my test several times to be sure it passes (the condition has to be random), the failure of this unit test will bleed into others. When the first `assoc()` test throws, the unit test will fail and Mocha will move on to the next. But there will usually still be some `assoc()` tests running in the background, which then also throw while Mocha is running some other test, therefore making it think those are failing too.

Shouldn't be a problem once this test passes, or is removed. But something to keep in mind if you decide to play around with this.

Avaq:
In name of lowering the number of PRs. This will probably never be merged.

buzzdecafe:
My apologies for the delay. I was taking a break from all these projects for a while and this one got missed. If you feel like this is a valuable contribution, please reopen and we'll try to be more responsive. Thanks

TheLudd:
I am just reading up on semigroups and believe I might be getting an understanding of them. How is it determined that this is what `concat` means for `Future`? Other types, such as [Identity](https://github.com/fantasyland/fantasy-land/blob/master/id.js#L12) from the fantasy-land repo require the inner type to also be a semigroup. Why could that not be the case here?

I am not saying that this is incorrect, just that I have no way of knowing i it is or not right now...

Avaq:
> we'll try to be more responsive -- @buzzdecafe 

Hey, no worries! I can image it wasn't very clear what to do with this PR.

> If you feel like this is a valuable contribution, please reopen -- @buzzdecafe 

We don't really need to reopen the PR, it was mostly here for discussion. I came to terms with the fact* that Future can't be a lawful Semigroup in JavaScript, or at least when implemented like this, to serve as a "race".. Though I wonder how many of the other Fantasy Land laws will not be obliged when testing with irregularly timed Futures.

Maybe we'll agree to ignore the issue of random time and call it "lawful enough", in which case reopening may be an option.

> How is it determined that this is what concat means for Future? -- @TheLudd 

I've been wondering the same thing. Once I saw that some Semigroup implementations simply delegate behaviour to contained Semigroups, I realized there is an infinity of thinkable ways to implement Semigroup lawfully. I choose this particular one because of its usefulness to async control.

TheLudd:
I was also confused about the constant delegation. But then I saw [these types from loop-recur](https://github.com/loop-recur/hostedjs/blob/master/v3/monoids.js) and realized that that there are some types where concat and emtpy are natrual constructs, array would be another one, perhaps object. However for any of the types currently in ramda-fantasy I see no such natural constructs at the moment. It seems to me they should all be delegating. However, that could also be achieved with `lift2(R.concat, a, b)`, so would there be any use?

Needles to say I am not clear on this =)

Avaq:
> [T]here are some types where concat and emtpy are nat[ur]al constructs

I think see what you mean: With some types it's obvious how `concat` should behave. In the cases that this is not obvious, perhaps it's a good idea not to implement it at all, because implementing it would involve businessy logic?

TheLudd:
Yes and businessy logic is to be avoided. I would not mind a function `race` however. Only question is if it should accept two Futures or a list o Futures. I don't have any races in the apps I build so I wouldn't know what is most suitable.

Avaq:
@TheLudd I've implemented two methods for "racing" in the [common Future related utilities that I use a lot](https://github.com/Avaq/node-server-skeleton/blob/master/src/util/future.js#L122-L187). Both methods take two Futures, and can be made to accept a list of Futures by simply passing them to `reduce`.

The first method (`race`) resolves with the first Future to resolve or reject. When passed to `reduce`, it'll give you a Future which resolves or rejects with the first resolution or rejection encountered.

The second method (`or`) resolves with the left Future, or if that rejects, with the right Future, or it rejects with the rejection reason of the right Future if both have rejected. When passed to `reduce`, you are guaranteed to receive the first Future, starting from the left, to resolve.

scott-christopher:
One practice used elsewhere (e.g. [Haskell's async][1] and [scalaz's Nondeterminism][2]) for modelling async data types is to have a separate wrapper type for handling non-deterministic behaviour while keeping the base async type's ordering linear and deterministic.

Things like the semigroup implementation, `race` and the current implementation of `Future::ap` would fall under the the non-deterministic type.

I suggested some time ago that I would revisit the implementation of `Future` but got distracted while exploring how hard it would be to do so via free monads. I can try to have a play with this again over the weekend and put together a proposal if it is seems like a viable option.

[1]: https://hackage.haskell.org/package/async
[2]: http://docs.typelevel.org/api/scalaz/nightly/index.html#scalaz.Nondeterminism

PR #84, opened 2015-12-24
Docs - first draft

scott-christopher:
It's here :smile: ... at least a first draft anyway.

I welcome plenty of feedback, as writing is definitely not a strong point of mine.

I've only captured the data types in this batch, but I would like to add more general docs on the different implemented specs at some point.

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,10 +3,6 @@ ramda-fantasy
 
 [Fantasy Land][1] compatible types for easy integration with [Ramda][2].
 
-
-[1]: https://github.com/fantasyland/fantasy-land
-[2]: https://github.com/ramda/ramda
-
 ## Project status
 This project is in alpha status. The implementation of the Fantasy Land spec should be *mostly*
 stable. Any methods outside of the Fantasy Land spec are subject to change. The types also have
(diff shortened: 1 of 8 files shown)


docs/Future.md
```diff
@@ -0,0 +1,108 @@
...
+#### `Future.reject :: e -> Future e a`
+Creates a `Future` instance that rejects with the provided value.
+
+#### `Future.cache :: Future e a -> Future e a`
+Creates a new `Future` instance from the provided `Future` instance, where the
+resolved or rejected value is cached for subsequent calls to `fork`.
+
+### Instance methods
+
+#### `future.fork :: Future e a ~> (e -> ()) -> (a -> ()) -> ()
```
davidchambers: We're missing a <code>`</code> at the end of this heading.

docs/IO.md
```diff
@@ -0,0 +1,79 @@
...
+
+#### `IO :: (() -> a) -> IO a`
+Constructs an `IO` instance that represents some action that may possibly have
+side effects.
+
+### Static methods
+
+#### `IO.runIO :: IO a -> a`
+Executes the action described by the given `IO` instance. This is also available
+as an instance method.
```
davidchambers: Should we document the instance method as well?

docs/Maybe.md
```diff
@@ -0,0 +1,120 @@
...
+plus1or0(safeDiv(42, 0)); // 0
+```
+
+Quite often it is useful transform the potential value of a `Maybe`, while
+deferring the extraction until later in the program. The two functions, `map`
+and `chain`, exist to support this behaviour. Both of these functions are
+somewhat similar in that neither can transform a `Nothing` into a `Just`, though
+`chain` is considered more powerful as it allows a function to transform a
+`Just` into a `Nothing`, while `map` can only transform the value contained
+within a `Just`.
```
davidchambers: Excellent description!

docs/Tuple.md
```diff
@@ -0,0 +1,71 @@
...
+
+Returns the first element of a given `Tuple`.
+
+#### `Tuple.snd :: Tuple a b -> b`
+
+Returns the second element of a given `Tuple`.
+
+### Instance methods
+
+#### `tuple.concat :: Semigroup a, Semigroup b => Tuple a b ~> Tuple a b -> Tuple a b`
```
davidchambers: Should we parenthesize `Semigroup a, Semigroup b`?

docs/Maybe.md
kedashoe: Quite often it is useful ***to*** transform...?

docs/Future.md
scott-christopher: :zap:

docs/IO.md
scott-christopher: Yep. :zap:

docs/Maybe.md
scott-christopher: :zap: updated to:
> It is quite often useful to transform the potential value of a `Maybe` while deferring the extraction until later in the program.

davidchambers:
This is terrific, Scott!

scott-christopher:
Thanks David.

My current thoughts:
* do the docs for the different types read consistently as a whole? 
* are there enough examples, do they make sense and do they actually help to explain the type?
* the type signatures in the references don't look great in Github's markdown rendering. Perhaps it might look nicer if the type signatures were separated onto the line beneath the method names?

davidchambers:
> the type signatures in the references don't look great in Github's markdown rendering. Perhaps it might look nicer if the type signatures were separated onto the line beneath the method names?

This might be a good idea. They could even be marked up as Haskell code blocks if Pygments doesn't complain about the `~>` and other "invalid" syntax.

docs/Future.md
```diff
@@ -0,0 +1,108 @@
...
+
+## Interaction
+
+Once a `Future` instance has been created, the various methods attached to the
+instance can be used to instruct further transformations to take place after
+the `Future` has been resolved or rejected. It is important to note that nothing
+is actually executed until the `fork` method is eventually called.
+
+The `map`, `ap` and `chain` functions can be used to transform resolved values
+of a `Promise` instance, while `chainReject` can be used to transform or recover
```
kedashoe: ... resolved values of a ***`Future`***... ?

scott-christopher: :zap:

docs/Tuple.md
scott-christopher: I don't mind either way. Probably a good idea to keep with the Haskell-like syntax.

scott-christopher:
I've split the type signatures out to their own line and annotated as `hs` as suggested. I think it's an improvement.

Perhaps we might be best publishing a templated version somewhere in the future.

README.md
```diff
@@ -16,16 +12,26 @@ not undergone thorough testing/use yet.
 
  Name       | Setoid  | Semigroup | Functor | Applicative | Monad | Comonad  | Foldable |
```
kedashoe: What about linking these column headers to their definitions in the Fantasy Land spec eg [Setoid](https://github.com/fantasyland/fantasy-land#setoid)

davidchambers:
> I've split the type signatures out to their own line and annotated as `hs` as suggested. I think it's an improvement.

I agree.

README.md
scott-christopher: Good idea. :zap:

davidchambers:
:deciduous_tree:

PR #85, opened 2015-12-29
Allow Ramda 0.19

jimf:
Update dependency on Ramda to work with 0.19.

### DIFF
--- a/package.json
+++ b/package.json
@@ -30,7 +30,7 @@
     "test": "mocha"
   },
   "dependencies": {
-    "ramda": ">=0.15.0 <=0.18.0"
+    "ramda": ">=0.15.0 <=0.19.0"
   },
   "devDependencies": {
     "jscs": "1.13.x",
--- a/test/future.test.js
+++ b/test/future.test.js
@@ -118,6 +118,8 @@ describe('Future', function() {
       };
     }
 
+    function noop() {}
+
     it('applies its function to the passed in future', function() {
       var f1 = Future.of(add(1));
       var result = f1.ap(Future.of(2));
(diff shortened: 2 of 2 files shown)


src/Future.js
```diff
@@ -33,7 +33,7 @@ Future.prototype.ap = function(m) {
 
   return new Future(function(rej, res) {
     var applyFn, val;
-    var doReject = R.once(rej);
+    var doReject = rej != null ? R.once(rej) : function() {};
```
scott-christopher: Is `null` sneaking into the `rej` argument via some code elsewhere in this type for this occur, or is it due to `null` being given to `fork`? If the latter, I'd prefer to not encourage `null` being accepted as argument values here and recommend the no-op function that's defaulted to here be used as the argument when `fork` is called instead.

Perhaps an alternative would be to create a `forkResolved :: (a -> ()) -> ()` for this use case, which only accepts the resolved callback.
```
Future.prototype.forkResolved = function(res) {
  return this.fork(function noop(){}, res);
};
```

jimf: `null` being passed to `fork`. I fully agree with your preference. I was just hesitant to introduce a potentially breaking change (though admittedly, nothing officially documents (afaik) that `fork` can accept `null` as a noop). I'll update.

scott-christopher:
Thanks @jimf. And sorry, I didn't realise the `null` values were being used in the tests.

:mushroom:

TheLudd:
What was the reason for changing `null` to `noop`? 
I think the `null` values was to force an error in the test when the future resolved to an error. If that happens now it will not be detected. If not `null`, then `assert.fail` should be used.

TheLudd:
Sorry, didn't notice the discussion on the outdated diffs. But please replace the `noop` with `assert.fail`. `null` is not really a valid argument to `fork` but it did serve a purpose in asserting the functionality in the tests.

scott-christopher:
If we're wanting to assert that the `reject` callback doesn't get called in these tests, then `assert.fail` is better than either a `noop` or `null` here.

PR #86, opened 2015-12-30
adds Either.isLeft/isRight

stoeffel:
How do you feel about adding a `Either.isLeft` and a `Either.isRight`?
Similarly to `Maybe.isJust` it can be called on an instance or as a static method.

Let me know if I need to change something.

### DIFF
--- a/test/either.test.js
+++ b/test/either.test.js
@@ -59,6 +59,27 @@ describe('Either', function() {
     jsv.assert(jsv.forall(eNatArb, fnNatArb, fnNatArb, eTest.associative));
   });
 
+  describe('checking for Left | Right', function() {
+    it('should allow the user to check if the instance is a Left', function() {
+      jsv.assert(jsv.forall(leftArb(jsv.nat), rightArb(jsv.nat), function(l, r) {
+        return l.isLeft === true && r.isLeft === false;
+      }));
+    });
+
+    it('should allow the user to check if the instance is a Right', function() {
+      jsv.assert(jsv.forall(leftArb(jsv.nat), rightArb(jsv.nat), function(l, r) {
+        return l.isRight === false && r.isRight === true;
+      }));
+    });
+
+    it('can check the type statically', function() {
+      jsv.assert(jsv.forall(leftArb(jsv.nat), rightArb(jsv.nat), function(l, r) {
+        return Either.isRight(l) === false && Either.isRight(r) === true &&
+               Either.isLeft(l) === true && Either.isLeft(r) === false;
+      }));
+    });
+  });
+
   describe('#bimap', function() {
 
     it('maps the first function over the left value', function() {
--- a/docs/Either.md
+++ b/docs/Either.md
@@ -74,6 +74,18 @@ the types of values contained in both `Left` and `Right`.
 Creates a pure instance of an `Either`. This is effectively a `Right`
 constructor.
 
+#### `Either.isLeft`
+```hs
+:: Either a b -> Boolean
+```
+Returns `true` if the given `Either` instance is a `Left`, otherwise `false`.
+
+#### `Either.isRight`
+```hs
+:: Either a b -> Boolean
+```
+Returns `true` if the given `Either` instance is a `Right`, otherwise `false`.
+
 ### Instance methods
 
 #### `either.map`
--- a/src/Either.js
+++ b/src/Either.js
@@ -24,6 +24,7 @@ Either.of = Either.prototype.of = function(value) {
 
 Either.prototype.chain = util.returnThis; // throw?
 
+
 Either.equals = Either.prototype.equals = util.getEquals(Either);
 
 Either.either = R.curry(function either(leftFn, rightFn, e) {
@@ -36,12 +37,23 @@ Either.either = R.curry(function either(leftFn, rightFn, e) {
   }
 });
 
+Either.isLeft = function(x) {
+  return x.isLeft;
+};
+
+Either.isRight = function(x) {
+  return x.isRight;
+};
+
 
 // Right
 function _Right(x) {
   this.value = x;
 }
 util.extend(_Right, Either);
+
+_Right.prototype.isRight = true;
+_Right.prototype.isLeft = false;
 
 _Right.prototype.map = function(fn) {
   return new _Right(fn(this.value));
@@ -77,6 +89,9 @@ function _Left(x) {
   this.value = x;
 }
 util.extend(_Left, Either);
+
+_Left.prototype.isLeft = true;
+_Left.prototype.isRight = false;
 
 _Left.prototype.ap = util.returnThis;
 


scott-christopher:
This would be good for the consistency with `Maybe`, however I'm not such a fan of `isJust` as well.

Do you have some examples of how `isLeft`/`isRight` would be used that aren't supported by the existing methods on `Either`? `Either.either` can handle extraction of both possible values and `Either.bimap` supports mapping of both possible values.

The slight concern I have with exposing methods like `isLeft`/`isRight` is that it tends to lead to the creation of partial functions, where only one code path is written for.

stoeffel:
> Do you have some examples of how isLeft/isRight would be used that aren't supported by the existing methods on Either? Either.either can handle extraction of both possible values and Either.bimap supports mapping of both possible values.

I'm working on a redux middleware to handle `Either` and I want to keep it generic. So I want to pass the checks for isLeft and isRight to the middleware (because it's handle different in i.e. folktale and ramda-fantasy) (see https://github.com/stoeffel/redux-either#usage). The middleware will dispatch either the value of `Left` as an `error` or the `Right` as the `payload` [redux-either#L21](https://github.com/stoeffel/redux-either/blob/master/src/index.js#L21)

scott-christopher:
So in that particular example:
```js
isLeft(action.payload)
      ? dispatch({ ...omit('payload')(action), error: either(action.payload) })
      : dispatch({ ...action, payload: either(action.payload) })
```
This could be modified to instead accept the equivalent of RF's `Either.either` or Folktale's `either.fold`:
```js
either(leftVal  => dispatch({ ...omit('payload')(action), error: leftVal }),
       rightVal => dispatch({ ...action, payload: rightVal }),
       action.payload);
```
and then
```js
const folktaleMW = applyMiddleware(eitherMiddleware((l, r, e) => e.fold(l, r)))(createStore)
// or
const ramdaFantasyMW = applyMiddleware(eitherMiddleware(Either.either))(createStore)
```

stoeffel:
Nice! Did just think about refactoring it to something like that too.

stoeffel:
Then the only reason to add this, would be to be consistent with Maybe.

davidchambers:
At Plaid, we often require `isLeft` and `isRight` at the gate between the sanctuary and existing code. For example:

```javascript
if (R.is(S.Left, either)) {
  process.stderr.write(String(either.value) + '\n');
  callback(either.value);
} else {
  callback(null, either.value);
}
```

`R.is` has proven problematic, so in plaid/sanctuary#105 we added `isLeft` and `isRight` functions, allowing `S.isLeft(either)`. Every Either value has `isLeft :: Boolean` and `isRight :: Boolean`, so one could instead write `either.isLeft` in this context.

Unless there's a better way to interface with existing callback-based and promise-based code, the issue raised in plaid/sanctuary#100 will affect ramda-fantasy users at some point, so it's worth considering `isLeft` and `isRight`.

/cc @michaelckelly

scott-christopher:
@davidchambers: Doesn't your example translate to the following?
```js
Either.either(
  leftX  => { process.stderr.write(String(leftX) + '\n'); return callback(leftX) },
  rightX => callback(null, rightX)
  either
);
```

The hesitation I have with including these is that it becomes possible for developers to forget the `else` case when checking one half of the possible types with `isLeft`/`isRight`, leading to partial functions.

davidchambers:
You're quite right, Scott, and I share your hesitation.

For some reason, though, applying `Either.either` to effectful, imperative functions feels unnatural, particularly when the functions have 10- or 20-line bodies rather than one- or two-line bodies.

scott-christopher:
You can always squint and pretend they're in some `IO`/`Eff` type. :P

scott-christopher:
I'm happy to include this if people see a genuine benefit from it, however it feels like we might be lacking something else if people feel a need for it.

Any thoughts from @TheLudd or @buzzdecafe?

buzzdecafe:
> Any thoughts from @TheLudd or @buzzdecafe?

Not much to add.  Since we have it on `Maybe` I don't see a reason *not* to have it for other types.

> it feels like we might be lacking something else if people feel a need for it.

Agreed. _ad hoc_ run-time type-checking feels pretty gross. I suspect that in javascript, that is probably as good as you can do. You might be able to do better with a compilation step; but if you are going to compile it, why would you write in javascript?

TheLudd:
I am thinking that perhaps including `isJust` et al was a mistake and that there are better ways of handling the situations where one might feel a need for such functions. Better solutions would be `Maybe.maybe` .`Either.either`, `reduce` and so forth.

I never user `isJust` these days.

stoeffel:
@TheLudd agreed. `either`, `maybe` and `reduce` are the better solution, but IMO sometimes it's just convenient to have `isLeft`, `isJust and the others.

TheLudd:
Ok. I will not give either +1 or -1 to this. I can certainly live without it but won't be super bothered by its inclusion.

stoeffel:
Either way I think it should be consistent with `Maybe`.

scott-christopher:
Could we move the methods onto the prototypes of the `Left` and `Right` so we don't have to perform `instanceof` tests?

```js
_Left.prototype.isLeft = R.T;
_Right.prototype.isLeft = R.F;
Either.isLeft = function(x) {
  return typeof x.isLeft === 'function' && x.isLeft();
}
// rinse/repeat for `isRight`
```

`Maybe.isJust` should probably be updated in a similar fashion too.

stoeffel:
Sure. will do that tomorrow.

buzzdecafe:
> Could we move the methods onto the prototypes of the Left and Right so we don't have to perform instanceof tests?

:+1:

there may be potential usefulness in these for pseudo-pattern-matching

davidchambers:
> ```javascript
> _Left.prototype.isLeft = R.T;
> _Right.prototype.isLeft = R.F;
> Either.isLeft = function(x) {
>   return typeof x.isLeft === 'function' && x.isLeft();
> }
> ```

This implementation implies the following type:

```haskell
Either.isLeft :: Any -> Boolean
```

I consider the following more appropriate:

```haskell
Either.isLeft :: Either a b -> Boolean
```

We could leave the behaviour of applying `Either.isLeft` to a non-Either value undefined, though I suggest throwing an exception with a descriptive message.

scott-christopher:
> This implementation implies the following type: `Either.isLeft :: Any -> Boolean`

I wasn't initially going to suggest the use of `typeof x.isLeft === 'function'`, but if `Maybe.isJust` is to be implemented the same way then it would also change from `Any -> Boolean` to `Maybe a -> Boolean` due to its current use of `instanceof`. I'd prefer `Maybe a -> Boolean`, but it would be a potentially breaking change.

davidchambers:
> I'd prefer `Maybe a -> Boolean`, but it would be a potentially breaking change.

It seems the best course of action is to create a separate pull request to update the type of `Maybe.isNothing` and `Maybe.isJust` to `Maybe a -> Boolean`, merge it (if others agree), then update this pull request to use `Either a b -> Boolean`.

stoeffel:
> Could we move the methods onto the prototypes of the Left and Right so we don't have to perform instanceof tests?

Done.

> I suggest throwing an exception with a descriptive message.

Did that as well and rebased the PR.

Let me know if there is anything else I need to change.
... and happy new year to you guys :fireworks:

test/either.test.js
```diff
@@ -59,6 +59,32 @@ describe('Either', function() {
     jsv.assert(jsv.forall(eNatArb, fnNatArb, fnNatArb, eTest.associative));
   });
 
+  describe('checking for Left | Right', function() {
```
scott-christopher: Would you mind updating these tests to use the jsverify arbitrary instances for `Either`? You can use the `leftArb` and `rightArb` functions to create `Left` and `Right` instances respectively.

```js
it('should allow the user to check if the instance is a Left', function() {
  jsv.assert(jsv.forall(leftArb(jsv.nat), rightArb(jsv.nat), function(l, r) {
    return l.isLeft() === true && r.isLeft() === false;
  }));
});
```

scott-christopher:
> @stoeffel ... and happy new year to you guys

And to you too :tada: :smile:

test/either.test.js
stoeffel: > Would you mind updating these tests to use the jsverify arbitrary instances for Either?

done

stoeffel:
I changed the tests to use jsv and I changed isLeft and isRight on the instance to be a boolean (like in #88)

test/either.test.js
```diff
@@ -59,6 +59,34 @@ describe('Either', function() {
...
+      }));
+    });
+
+    it('should allow the user to check if the instance is a Right', function() {
+      jsv.assert(jsv.forall(leftArb(jsv.nat), rightArb(jsv.nat), function(l, r) {
+        return l.isRight === false && r.isRight === true;
+      }));
+    });
+
+    it('can check the type statically', function() {
```
scott-christopher: This test should be updated to use `jsv` too.
```js
it('can check the type statically', function() {
  jsv.assert(jsv.forall(leftArb(jsv.nat), rightArb(jsv.nat), function(l, r) {
    return Either.isRight(l) === false && Either.isRight(r) === true &&
           Either.isLeft(l) === true && Either.isLeft(r) === false;
  }));
});
```

stoeffel: sure. will do.

scott-christopher:
Thanks for persisting with this, @stoeffel. Much appreciated.

scott-christopher:
:mushroom:

stoeffel:
Awesome. Thanks

PR #87, opened 2015-12-31
Fix test semantics of reject for Future#ap

jimf:
Replace noops with the more correct `assert.fail` to ensure test failure
in the event that the reject callback is run.

See #85 for discussion.


scott-christopher:
Thanks @jimf

jimf:
:beers:

PR #88, opened 2016-01-01
isJust/isNothing without instanceof check

stoeffel:
- throwing an exception if not a Maybe is passed

refs #86

### DIFF
--- a/src/Maybe.js
+++ b/src/Maybe.js
@@ -11,8 +11,14 @@ function _Just(x) {
 }
 util.extend(_Just, Maybe);
 
+_Just.prototype.isJust = true;
+_Just.prototype.isNothing = false;
+
 function _Nothing() {}
 util.extend(_Nothing, Maybe);
+
+_Nothing.prototype.isNothing = true;
+_Nothing.prototype.isJust = false;
 
 var _nothing = new _Nothing();
 
@@ -29,11 +35,11 @@ Maybe.of = Maybe.Just;
 Maybe.prototype.of = Maybe.Just;
 
 Maybe.isJust = function(x) {
-  return x instanceof _Just;
+  return x.isJust;
 };
 
 Maybe.isNothing = function(x) {
-  return x === _nothing;
+  return x.isNothing;
 };
 
 Maybe.maybe = R.curry(function(nothingVal, justFn, m) {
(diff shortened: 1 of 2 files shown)


src/Maybe.js
```diff
@@ -11,9 +11,15 @@ function _Just(x) {
 }
 util.extend(_Just, Maybe);
 
+_Just.prototype.isJust = R.T;
+_Just.prototype.isNothing = R.F;
```
davidchambers: Why not the following?

```javascript
_Just.prototype.isJust = true;
_Just.prototype.isNothing = false;
```

stoeffel: Would work as well. But I think it's more intuitive when `isJust`, `isNothing`, `isLeft` and `isRight` are functions.
But I could live with both.

scott-christopher: I think I'd prefer them both to have static properties on their prototypes rather than functions too (apologies for my earlier suggestion in #86 for using `R.T`/`R.F`). We still have the functions available on `Maybe`/`Either` for people to use where a function is more appropriate.

stoeffel:
I changed `isJust` and `isNothing` on the instance to be a boolean. :checkered_flag:

src/Maybe.js
```diff
@@ -29,11 +35,19 @@ Maybe.of = Maybe.Just;
 Maybe.prototype.of = Maybe.Just;
 
 Maybe.isJust = function(x) {
-  return x instanceof _Just;
+  if (typeof x.isJust !== 'boolean') {
```
scott-christopher: I think the `typeof` test is no longer necessary here (or in `isNothing`) and we can just return `x.isJust === true`;

stoeffel: Then we need to change the signature to `Maybe.isJust :: Any -> boolean`

scott-christopher: We can still keep the types signature to specify that only `Maybe` instances are accepted. It's just a matter of whether we throw in the implementation if something other than a `Maybe` instance is given.

If we're to add type checking to functions, we will need to approach it more holistically such as the suggestion in #89.

stoeffel: convinced. will drop it.

scott-christopher: Would you mind doing the same in the `Either` PR #86? Then I think we're in a good position for both of them. :)

stoeffel: Sure will do.

stoeffel: and done.

scott-christopher:
Thanks again, @stoeffel :mushroom:

Issue #91, opened 2016-01-04
Plans for State and Writer


achernyak:
Are there any plans to implement State and Writer?

scott-christopher:
I've got a `State` implementation locally that I can push up a PR for. Nothing for `Writer`, but I'm happy to look into it too.

CrossEye:
While a Writer seems obviously useful, I'm quite curious to see what a State implementation buys JS users.

joneshf:
`Writer` already exists: https://github.com/ramda/ramda-fantasy/blob/master/src/Tuple.js

scott-christopher:
I've pushed a branch up for `State` if people are interested in having a look. I'll hold off raising a PR until I've finished the docs for it sometime in the next day or so. https://github.com/ramda/ramda-fantasy/compare/master...scott-christopher:state

@CrossEye: I've included the cliché example of a deterministic PRNG using `State` in the docs which might spark some ideas for uses. https://github.com/scott-christopher/ramda-fantasy/blob/state/docs/State.md

The reason I implemented it initially was to have access to a "pure" monad that could be swapped in place of `IO`/`Future` as the interpreter while testing some free monad logic.

CrossEye:
@scott-christopher:
> I've included the cliché example of a deterministic PRNG using State in the docs which might spark some ideas for uses

My question was how these might be of use in Javascript, where mutable local states are a dime a dozen.  Having a special construct that makes it possible to deal with local states is important in languages that prohibit or discourage other forms of state.  But in JS, I'm not sure I see the advantage of the State monad over say, this:

```js
const pnrg = seed => () => {
  seed = (seed * 1103515245 + 12345) & 0x7FFFFFFF;
  return (seed >>> 16) / 0x7FFF;
};

const rng = pnrg(42);
rng(); //=> 0.5823236793115024
rng(); //=> 0.5198217719046602

var rollDie = R.map(n => 1 + ((n * 6) >>> 0), rng);
R.times(rollDie, 10); //=> [6, 5, 5, 5, 4, 1, 1, 2, 2, 6]
```

[...]

PR #92, opened 2016-01-06
Adds State type

scott-christopher:
As per #91, this change introduces the `State` type and `StateT` transformer.

Docs can be viewed here: https://github.com/scott-christopher/ramda-fantasy/blob/state/docs/State.md

### DIFF
--- a/docs/State.md
+++ b/docs/State.md
@@ -1,0 +1,246 @@
+# State
+
+The `State` type can be used to store some state along with a computation.
+
+## Construction
+
+`State` instances should be obtained via the `get`, `gets`, `put`, and `modify`
+static properties on the `State` object. These instances describe the different
+ways to access and modify the stateful computation that will eventually be
+evaluated and are described in further detail below.
+
+A `State` transformer is also available via `State.T` which can be used to
+extend some monad with stateful behaviour.
+
+## Interaction
+
+`State` instances are primarily interacted with and composed via the `chain`
+method of the various static instances available on the `State` type object. To
+access the current state, `State.get` is a static instance that can be used to
+provide the state to the `chain`, `ap` and `map`. Similarly, `State.gets(fn)`
+can provide the state transformed by the provided `fn` function. To change the
+state of the computation, `State.put(newValue)` can be used to replace the
+existing state with `newValue`. The current state can also be transformed by
+providing a transformation function to `State.modify(transformFn)`.
+
+Once a `State` instance is defined, an initial seed state can be provided to
+either the `eval` or `exec` methods to evaluate the computation and return the
+result of the computation or the final state, respectively. Alternatively, the
+`run` method can be called to return both the result and the final state within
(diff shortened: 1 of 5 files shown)


docs/State.md
```diff
@@ -0,0 +1,238 @@
...
+  return State.put(newSeed).map(_ => randVal);
+});
+
+rng.eval(42); // From initial seed of 42: 0.5823236793115024
+rng.eval(42); // Repeating produces the same value: 0.5823236793115024
+rng.exec(42); // `exec` returns the next seed: 1250496027
+rng.eval(1250496027); // A different seed: 0.5198217719046602
+
+// Chain together to pass the new seed to the next RNG
+const pair = rng => rng.chain(a => rng.chain(b => State.of(Tuple(a, b))));
```
davidchambers: I'd find a type signature very helpful here. :)

docs/State.md
```diff
@@ -0,0 +1,238 @@
...
+```
+Constructs a `State` instance that represent a pure computation from some state
+to a new state and a result. Note this constructor requires the given function
+to return an `Identity` instance. It is generally recommended to use the static
+properties and methods provided on the `State` object rather than using this
+constructor.
+
+#### `State.T`
+```hs
+:: Monad m => { of: a -> m a } -> (s -> m (Tuple a s)) -> StateT s m a
```
davidchambers: How about `{ of :: a -> m a }` rather than `{ of: a -> m a }`?

test/state.test.js
```diff
@@ -0,0 +1,113 @@
...
+function stateArb(evalArb, execArb) {
+  return jsv.fn(tupleArb(evalArb, execArb)).smap(
+    function (f) { return State(R.compose(Identity, f)); },
+    function (s) { return s.run; }
+  );
+}
+
+function identityArb(valueArb) {
+  return valueArb.smap(
+    function (a) { return Identity(a); },
```
davidchambers: Could we provide `Identity` here directly?

scott-christopher: Yep, good catch. :zap:

docs/State.md
scott-christopher: Sure. :zap:

docs/State.md
scott-christopher: I've scattered a few more type signatures through this example. :zap:

davidchambers: Thanks, Scott! Very helpful indeed. :)

docs/State.md
```diff
@@ -0,0 +1,246 @@
...
+
+#### `StateT.lift`
+```hs
+:: Moand m => m a -> StateT s m a
+```
+Lifts the given monad into a `StateT` instance.
+
+### Instance methods
+
+#### `state.run`
```
davidchambers: What do you think of using `State#run` to indicate an instance method rather than `state.run`?

PR #93, opened 2016-01-12
Update "Setoid" column in "Available types" table

rpominov:
As far as I understand Setoid implementation was removed for some types, but table in Readme haven't been updated.

Not quite sure about Reader though, seems like it still has some Setoid impl in monad transformer, but I yet to understand transformers (not sure how they work).


scott-christopher:
:mushroom: 

Thanks @rpominov. Any type that is effectively a function probably shouldn't attempt to implement `Setoid`. I've raised #94 to remove `equals` from `ReaderT` too.

PR #96, opened 2016-01-24
Fix function name in IO docs

jimf:


scott-christopher:
Oops :smile: Thanks @jimf.

PR #98, opened 2016-01-26
Tree-shake ramda by cherry-picking requires

jimf:
Slim down browserify-based environments by cherry-picking the needed
requirements from Ramda rather than pulling in the whole thing.

Style should be consistent with Ramda-proper's source.

### DIFF
--- a/src/Either.js
+++ b/src/Either.js
@@ -1,4 +1,5 @@
-var R = require('ramda');
+var curry = require('ramda/src/curry');
+var toString = require('ramda/src/toString');
 
 var util = require('./internal/util');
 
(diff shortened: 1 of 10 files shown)


buzzdecafe:
makes sense. :cow2:

TheLudd:
How does this hold up using browserify? If I use browserify with ramda-fantasy and ramda when it is packaged this way, will it be the same size as before? Can I browserify ramda into a standalone bundle and ramda-fantasy into its own and still have the output no bigger than before?

jimf:
@TheLudd I believe your questions are dependent on both the version of browserify you're using, as well as the version of npm. However, my understanding is that by doing it this way, the resulting build will never be larger, and in most cases, it will be smaller. I'd encourage you to test it though, particularly if those use cases describe your general build process.

flintinatux:
👍  this. Importing all of `ramda` in front-end makes for huge builds.

buzzdecafe:
sorry for the *long* delay. are you interested in repairing conflicts? i promise a prompt merge this time

buzzdecafe:
#119 seems to have done this -- gonna close this one and merge that one

PR #101, opened 2016-02-21
fixes #100; rename _Just, _Nothing to Just, Nothing resp.

### LINKED ISSUE #100: Maybe toString gives away internals
No big deal, but perhaps worth raising here.

Noticed the toString on Maybe is displaying _Just rather than Just. Haven't checked other Types to see if they exhibit the same behaviour. It would be cleaner if it returned a toString of what the user is workiing with - Just and Nothing in this case.

```js
const nestedIceCream = Maybe(Maybe('Ice cream'));
console.log(nestedIceCream);
// _Just { value: _Just { value: 'Ice cream' } }
```

Thoughts?

buzzdecafe:


davidchambers:
:deciduous_tree:

PR #103, opened 2016-02-24
fixes #102, keep Nothing from getting mutated

### LINKED ISSUE #102: _nothing may leak, get mutated
```js
var n = Maybe.Nothing();
n.x = "mutated";
Maybe.Nothing().x; // "mutated"
```

this seems bad. I think we should rethink how `Nothing` is implemented. It relies too much upon the good behavior of the developer.

buzzdecafe:


src/Maybe.js
```diff
@@ -20,7 +20,7 @@ util.extend(Nothing, Maybe);
 Nothing.prototype.isNothing = true;
 Nothing.prototype.isJust = false;
 
-var _nothing = new Nothing();
+var _nothing = Object.freeze(new Nothing());
```
davidchambers: Which browsers does this project target? Is `Object.freeze` available in all of them?

It seems simpler to me to update the implementation of `Maybe.Nothing`:

```javascript
Maybe.Nothing = function() {
  return new Nothing();
};
```

buzzdecafe: > Is Object.freeze available in all of them?

I considered that. This looks good enough to me: http://caniuse.com/#search=freeze

i like the idea that there is one Nothing. Many Nothings feels pretty strange. (On reflection, *one* Nothing is a little strange.) But that is another possible solution.

buzzdecafe:
closing this; i think we may need a bigger rewrite to address this issue

Issue #106, opened 2016-02-28
Bower reference Ramda versioning


robjens:
The `package.json` shows a version range up to `0.20` but `bower.json` is fixed at `0.14`. ?
Is there a particular reason holding back the dependency for bower RF to be in line with `0.19.x` Ramda?

```
Unable to find a suitable version for ramda, please choose one:
    1) ramda#^0.14.0 which resolved to 0.14.0 and is required by ramda-fantasy#0.4.1
    2) ramda#^0.19.1 which resolved to 0.19.1 and is required by my-project
```

Cheers

buzzdecafe:
no reason, just a miss

robjens:
Ah, good. Or at least that was what I was hoping for. Lightning fast response by the way,  would an minor update be this fast as well? :)

buzzdecafe:
i'd be grateful for a PR :-)

robjens:
Ok sure... As I assume there won't be any/much conflicts. Just have to see if bower supports ranges as well and try it out locally first, then do a PR. This might require a new tagged release though, I'm not really sure but bower seems to kick-off on that, you happen to know? Anyway you can use the SHA to install a latest master commit as well, as opposed to tagged (snapshots) so we'll see.

buzzdecafe:
> As I assume there won't be any/much conflicts. 

probably a risky assumption, since ramda is still pre-1.0 and has undegones a lot of api changes in that span. ramda 0.9.1 was added as a dependency here: https://github.com/ramda/ramda-fantasy/commit/1ad9f7fe7de314f8bb6f1dfe10f60667e4507660 

probably not safe to rely on anything before that at least

robjens:
Oh.. my bad. I hadn't realized that part, actually missed the dependency range in package.json. Sorry.

PR #107, opened 2016-02-28
bower.json sugar for semver and recommended main

robjens:
This should work. Using sugar/2.0 additional syntax of `0.1.x` over `~1.0` as it is less clear imho, should work as well. Also added `main` to bower.json, recommended but not used directly though - build tools may in fact (and e.g. bower-requirejs likes it a lot too since this is an unambiguous application end-point. 
#106

### DIFF
--- a/bower.json
+++ b/bower.json
@@ -12,6 +12,7 @@
       "name": "Ludwig Magnusson"
     }
   ],
+  "main": "index.js",
   "homepage": "https://www.github.com/ramda/ramda-fantasy",
   "license": "MIT",
   "moduleType": [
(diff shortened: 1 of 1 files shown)


TheLudd:
This cant work. ramda-fantasy is not compatible with ramda versions lower than `0.14.0`

bower.json
```diff
@@ -29,7 +30,7 @@
     "type"
   ],
   "dependencies": {
-    "ramda": "^0.14.0"
+    "ramda": "0.1.x"
```
davidchambers: Should we change this to `"0.19.x"`?

TheLudd: Use the same range as in package.json.

robjens: Ah, my bad. So you corrected this already yes? I'm a tad unfamiliar with the whole interface here.

davidchambers: You could commit a change on your local __robjens-1__ branch and push the branch to GitHub to update this pull request.

buzzdecafe:
superseded by subsequent commits

Issue #105, opened 2016-02-25
Take Sanctuary


buzzdecafe:
Should we merge RF and Sanctuary?
What would a merger of RF an Sanctuary look like?

davidchambers:
I have quite a few thoughts on this matter. I hope to put them into writing later in the day. :)

buzzdecafe:
For me it boils down to:

* much of the work is duplicated; and 
* the goals of the two projects overlap; and
* RF was a learning project for me, now i don't devote much time to it;
* I have an idea I want to work on that probably fits better in sanctuary than RF.

> I have quite a few thoughts on this matter.

I don't doubt it

> I hope to put them into writing later in the day. :)

looking forward to it

joneshf:
:+1:

davidchambers:
Sanctuary and ramda-fantasy have much in common. Both libraries provide algebraic data types and functions for operating on values of those types.

The projects came into existence for different reasons: ramda-fantasy as a playground for types; Sanctuary to address limitations imposed by our decision not to have Ramda functions take or return values of Ramda-defined types.

Sanctuary's emphasis on correctness manifests in two ways. First, it means `head` has type `[a] -> Maybe a` rather than `[a] -> a`. Secondly, function contracts are enforced at run time. Run-time type checking is invaluable during development, so eventually Sanctuary will provide much of the Ramda API.

It would be nice to avoid duplicated effort (@scott-christopher has made significant contributions to both projects). How might this work? There are several points to consider:

  - __Name.__ In my view Ramda and Sanctuary are both excellent names. I also like the name ramda-fantasy, but others are not so happy with it (see #99). If this project and Sanctuary are to merge, Sanctuary seems the more appropriate name for the combined project: we leverage Fantasy Land to create a refuge from unsafe JavaScript.

  - __Organization.__ Sanctuary is currently in the __plaid__ namespace on GitHub. This would not be an appropriate location. Perhaps the various Sanctuary projects should live in the __ramda__ namespace. This is something we'd need to discuss with people at Plaid.

  - __Tools.__ The tooling plays a part in making contributing to Sanctuary enjoyable. There's one source file and one test file. No build script. No JSDoc. Comments in the source code flow through the readme to generate the [website][1]. Code examples are never out of date as they're tested with every commit. Full branch coverage is a requirement. These are all things I wish to preserve.


[1]: https://github.com/plaid/sanctuary-site/pull/2

scott-christopher:
[...]

PR #111, opened 2016-06-06
getOrElse instance method on either

wennergr:
:: Either a b ~> b -> b

Returns the right value if the instance is a `Right`, otherwise the provided
default value will be returned.

### DIFF
--- a/docs/Either.md
+++ b/docs/Either.md
@@ -135,6 +135,13 @@ the instance is a `Left`, or the second function if the instance is a `Right`.
 Determines whether the provided value is equal to this instance, ensuring both
 are of the same constructor and both contain equal values.
 
+#### `either.getOrElse`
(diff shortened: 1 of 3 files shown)


davidchambers:
How about `fromEither :: b -> Either a b -> b` instead, as proposed in sanctuary-js/sanctuary#130? This would be consistent with [`S.fromMaybe`][1]. Consistency between ramda-fantasy and Sanctuary seems valuable given the [goal][2] to merge ramda-fantasy and sanctuary-types (once the latter exists).


[1]: http://sanctuary.js.org/#fromMaybe
[2]: https://github.com/ramda/ramda-fantasy/issues/105

buzzdecafe:
closing in favor of `fromEither` as @davidchambers suggested

PR #116, opened 2016-08-08
Ramda version bump

kujon:


buzzdecafe:
thanks

kujon:
np :) any chance of publishing soon?

buzzdecafe:
hmmm i don't even recall what's new. is there something you need particularly?

kujon:
version mismatch in ramda and ramda-fantasy makes module bundlers bundle 2 versions of ramda together, thus increasing the size of js file.

buzzdecafe:
i can see that. i'll try to publish soon

tusharmath:
Have you considered https://github.com/semantic-release/semantic-release, you could automate the release process completely.

kujon:
@tusharmath using it ATM, it's awesome. The problem is when I bump Ramda version in my project and ramda-fantasy still references the older older version, I end up with 2 Ramdas in my bundle.

tusharmath:
@kujon I think you should do a `npm ddp` before creating a bundle — https://docs.npmjs.com/cli/dedupe

kujon:
@tusharmath would it affect the outcome if RF is `>=0.15.0 <=0.21.x` and my project is `0.22.x`?

buzzdecafe:
maybe we should just change package.json to `>=0.15.0`

tusharmath:
@kujon unfortunately in that case you will have two copies in the bundle.
@buzzdecafe greenkeeper is a good option together with https://github.com/semantic-release/semantic-release.

buzzdecafe:
interesting stuff @tusharmath. thanks, i'll look into it

kujon:
> maybe we should just change package.json to >=0.15.0
Didn't want to do it myself, but if we can assume RF will work with future iterations of ramda, then it's grand 👌

buzzdecafe:
> if we can assume RF will work with future iterations of ramda

i think the big api swings are over. one big forthcoming issue is the reduction of placeholder support, but i don't think that would affect RF. So I'd be content to leave the interval half open [0.15.0, &infin;)

buzzdecafe:
just published 0.6.1 which has no upper limit on R version

PR #119, opened 2016-08-19
Fixed the way Ramda is required

raqystyle:
This commit comes from my need to use Maybe only as I do not need the rest. Thus,
I do `require('ramda-fantasy/src/Maybe')`. But the problem with the old way of doing require in `ramda-fantasy` is that tree shaking can not be performed.
Although only three functions are used in this module from `ramda`, the whole library (`ramda`) is included in my build through `ramda-fantasy`.

I use rollup in order to build my code as it can perform tree shaking. Thus, the features that are not required in my code are not included in build. However, dependencies
also need to include the only parts they need.

### DIFF
--- a/src/Either.js
+++ b/src/Either.js
@@ -1,4 +1,5 @@
-var R = require('ramda');
+var curry = require('ramda/src/curry');
+var toString = require('ramda/src/toString');
 
 var util = require('./internal/util');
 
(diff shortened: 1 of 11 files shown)


buzzdecafe:
this would supersede #98 

lgtm :dromedary_camel: 

if there're no objections i'd like to merge this and publish 0.7 soon

PR #125, opened 2016-09-08
Implement FantasyLand ChainRec

### LINKED ISSUE #124: Implement FantasyLand ChainRec
As of [fantasyland/fantasy-land#152](https://github.com/fantasyland/fantasy-land/pull/152) implement [ChainRec](https://github.com/fantasyland/fantasy-land#chainrec) specification.

[MonadRec instance for either](https://github.com/purescript/purescript-tailrec/blob/master/src/Control/Monad/Rec/Class.purs#L89-L95) in purescript might be usefull

Here is [implementation](https://github.com/safareli/free/blob/e5fbbde2c50a1ec61968ec7c529aa3bb8ea51841/src/tailRec.js#L25-L98) of ChainRec for `folktale/data.task` with [tests](https://github.com/safareli/free/blob/e5fbbde2c50a1ec61968ec7c529aa3bb8ea51841/test/tailRecM.js) myght be usefull for `Future` (it's not quite correct as it is not handling cleanup stuff of the data.task but without that it works well)

I would create PR when have time.

safareli:
Implemented ChainRec for:
* [x] Either
* [x] Identity
* [x] Maybe
* [x] Future
* [x] IO

Not implemented for:
* Reader, ReaderT
* State, StateT

(As I don't quite understand Transformers, may create other PR when I get them)

Fixes #124


safareli:
Hey @buzzdecafe now you can do a review 🎉

buzzdecafe:
thanks -- there's a lot for me to digest here. I don't know how much time I have to review this weekend. I hope some other contributors may be able to have a look?

src/Either.js
```diff
@@ -70,6 +70,23 @@ _Right.prototype.chain = function(f) {
...
+  var state = util.chainRecNext(i);
+  while (state.done === false) {
+    state = Either.either(
+      function(v) {
+        return { done: true, value: v, isLeft: true };
+      },
+      function(v) {
+        return v;
+      },
+      f(util.chainRecNext, util.chainRecDone, state.value)
```
scott-christopher: We could make use of the existing `Either` to handle `next` and `done` here.

I think something like this should work:
```js
Either.chainRec = Either.prototype.chainRec = function(f, i) {
  var updateState = Either.either(
    compose(Right, Left),
    Either.either(compose(Left, Right), compose(Right, Right))
  );
  var state = Left(Right(i));
  while (Either.isLeft(state)) {
    state = updateState(f(Left, Right, state.value.value));
  }
  return state.value;
};
```

src/IO.js
```diff
@@ -21,6 +23,17 @@ IO.prototype.chain = function(f) {
   });
 };
 
+//chainRec
+IO.chainRec = IO.prototype.chainRec = function(f, i) {
+  return new IO(function() {
+    var state = util.chainRecNext(i);
+    while (state.done === false) {
+      state = f(util.chainRecNext, util.chainRecDone, state.value).fn();
```
scott-christopher: Likewise here, we could use `Left` and `Right` instead of `chainRecNext` and `chainRecDone`, along with `Either.isLeft(state)` instead of `state.done === false`.

src/Future.js
```diff
@@ -82,6 +84,79 @@ Future.prototype.chain = function(f) {  // Sorella's:
...
+  if (isSync) {
+    return { isSync: true };
+  } else {
+    isSync = true;
+    return { isSync: false };
+  }
+};
+
+// chainRec
+Future.chainRec = Future.prototype.chainRec = function(f, i) {
```
scott-christopher: There are a lot of moving parts in here which is proving to be a bit difficult for me to follow. I think perhaps some comments in the code could help out a lot here.

Alternatively, the implementation of the MonadRec instance for Aff translates quite nicely.
```js
// Heavily influenced by the Aff MonadRec instance
// https://github.com/slamdata/purescript-aff
function tailRecM(f, a) {
  var UNRESOLVED = 0,
      SYNC       = 1,
      ASYNC      = 2;

  return Future(function(reject, resolve) {
    function go(acc) {
      var result, status;
      while (true) {
        status = UNRESOLVED;
        f(Left, Right, acc).fork(reject, function _resolve(v) {
          if (status === UNRESOLVED) {
            // If this callback has been called while the status is still
            // UNRESOLVED then this has been a synchronous call. If this
            // were asynchronous, then the status would have been updated
            // below before this callback was invoked.
            status = SYNC;
            // Store the result for further synchronous processing.
            result = v;
          } else {
            // Otherwise it is safe to call recursively due to async callback.
            Either.either(go, resolve, v);
          }
        });

        // status will be SYNC if the above callback function has already been invoked.
        if (status === SYNC) {
          if (Either.isLeft(result)) {
            // update the value for the next loop and continue
            acc = result.value;
            continue;
          } else {
            // otherwise we're done
            resolve(result.value);
          }
        } else {
          // If the status hasn't been updated to SYNC yet, it must be ASYNC.
          status = ASYNC;
        }
        // async futures will exit the while loop here
        return;
      }
    }
    return go(a);
  });
}
```

scott-christopher:
This is great work @safareli! I can take a look into adding support for State/T and Reader/T.

src/Future.js
safareli: I have only looked at [implementation for Eff](https://github.com/purescript/purescript-tailrec/blob/master/src/Control/Monad/Rec/Class.purs#L97-L112) but I don't use and know purescript and it was too criptic for me.

The implementation you have linked and ported is nicer! Will update it 🌴

safareli:
@scott-christopher thanks, that would be awesome!

src/Either.js
safareli: I have updated my implementation but I think it's better to use custom structure for representing state instead of Either and it's also simpler to understand.

buzzdecafe:
apart from some detail-y stuff in the tests, LGTM

safareli:
@buzzdecafe I think you have accidentally closed it :p

buzzdecafe:
oops, sorry. it's been a long few days

test/future.test.js
```diff
@@ -65,7 +100,51 @@ describe('Future', function() {
...
+      var x = 1;
+      var initial = [x];
+      var next = function(a) {
+        return Future.of(a.concat([x]));
+      };
+      assert.equal(true, cTest.iface(Future.of(1)));
+      return cTest.equivalence(Future, predicate, done, next, initial);
+    });
+
+    it('sync and async Futures', function() {
```
safareli: fixed, it tests if it works when mixing sync and async futures

test/either.test.js
```diff
@@ -49,6 +49,48 @@ describe('Either', function() {
...
+      assert.equal(true, Either.of('DONE').equals(Either.chainRec(function(next, done, n) {
+        if (n === 0) {
+          return Either.of(done('DONE'));
+        } else {
+          return Either.of(next(n - 1));
+        }
+      }, 100000)));
+    });
+
+    it('fail Immediately', function() {
```
safareli: fixed

buzzdecafe:
:dromedary_camel:

safareli:
sweet 🍓

safareli:
Could we bump version?

scott-christopher:
v0.7.0 has been published.

PR #129, opened 2016-10-01
Umd

### LINKED ISSUE #43: needs UMD wrapper
I am transpiling my es6 to AMD modules with Babel. I may take a shot at putting this lib in the UMD wrapper so it will work with both flavors of modules.

buzzdecafe:
Fixes #43


buzzdecafe:
This exposes a global variable `RF` in browser environments

PR #131, opened 2016-10-20
make ramda-fantasy fantasy-land@1.0.x compatible

safareli:


safareli:
currently [this](https://github.com/safareli/free/blob/f49f78a401b16bfea3bb682398234fc1f36b0a71/test/lib/index.js#L8-L15) is what i'm doing

buzzdecafe:
this looks fine to me, although it does make me wonder again what the purpose of two separate libs (S & RF) is.  See #105

src/Either.js
```diff
@@ -125,5 +125,10 @@ Either.Left = function(value) {
   return new _Left(value);
 };
 
+require('./internal/fl-patch.js')([
```
davidchambers: For consistency let's drop `.js` from the path.

src/internal/fl-patch.js
```diff
@@ -0,0 +1,22 @@
+var fl = require('./fl.js');
+var fixedAp = function(f) {
+  return f.ap(this);
+};
+
+var patch = function(obj){
+  return Object.keys(fl).forEach(function(key) {
+    if (typeof obj[key] === 'function') {
+      if (key === 'ap') {
+        obj[fl[key]] = fixedAp;
```
davidchambers: This doesn't feel right to me. What would be the consequences of removing this exception?

package.json
```diff
@@ -30,7 +30,8 @@
     "test": "mocha"
   },
   "dependencies": {
-    "ramda": ">=0.15.0"
+    "ramda": ">=0.15.0",
+    "sanctuary-type-classes": "0.1.0"
```
davidchambers: I just published v0.2.0. :)

buzzdecafe: congrats

src/internal/fl-patch.js
safareli: removing this means we should revert `ap` of all structures

src/State.js
```diff
@@ -24,56 +25,59 @@ function T(M) {
...
   };
   StateT.prototype.ap = util.deriveAp(StateT);
   StateT.prototype.map = util.deriveMap(StateT);
   StateT.tailRec = curry(function(stepFn, init) {
     return StateT(function(s) {
       return M.tailRec(function(t) {
-        return stepFn(Tuple.fst(t))._run(Tuple.snd(t)).chain(function (t_) {
-          return M.of(Tuple.fst(t_).bimap(
+        return Z.chain(function (t_) {
+          return Z.of(M,Z.bimap(
```
davidchambers: `s/M,Z/M, Z/`

src/internal/fl-patch.js
davidchambers: Change the way the existing `ap` methods behave, do you mean? I think we should do so. Having an `ap` method with the old behaviour is confusing.

davidchambers:
We might as well go straight to `fantasy-land@2.0.0` now. :)

package.json
```diff
@@ -30,7 +30,8 @@
     "test": "mocha"
   },
   "dependencies": {
-    "ramda": ">=0.15.0"
+    "ramda": ">=0.15.0",
+    "sanctuary-type-classes": "0.2.0"
```
davidchambers: I just published v0.3.0. :)

scott-christopher:
Do we want this PR to land before we wrap up with a final release?

davidchambers:
> Do we want this PR to land before we wrap up with a final release?

Yes, I think so.

safareli:
`ap` needs to be redefined of all structures and S-type-classes should be updated, for this pr to be merged. I think i might have some time this weekend for it.

Issue #132, opened 2016-10-24
Discussion: Support toMaybe for Maybe Monad


martinlechner1:
I miss good way to wrap nullable values in the Maybe Monad. 
Option Types nowadays usually provide a static constructor ofNullable, which returns Just(x) or Nothing() depending on the value given.
I don't want to use Maybe(x) directly, since I want to show more explicitly that my wrapped value is nullable.
I could implement ofNullable if there is an agreement to support this.

scott-christopher:
I'd be in favour of moving the current logic from the constructor to a static method.

davidchambers:
We could name the function `toMaybe` to match [`S.toMaybe`][1].


[1]: https://sanctuary.js.org/#toMaybe

TheLudd:
I would prefer a function `nothingIf`

```
nothingIf:: (a -> Boolean) -> a -> Maybe[a]
```
Now you can do `nothingIf(isNil)` and you have your `Maybe` factory.
There could also be a `justIf` function that works the other way around and creates a `Nothing` if the predicate function evaluates to `false`.

davidchambers:
I don't see a need for `nothingIf`, @TheLudd. We can use [`filter`][1] or [`filterM`][2] instead:

```javascript
const S = require('sanctuary');
const Z = require('sanctuary-type-classes');

Z.filter(S.even, Just(41));  // => Nothing
Z.filter(S.even, Just(42));  // => Just(42)
Z.filterM(S.even, Just(41));  // => Nothing
Z.filterM(S.even, Just(42));  // => Just(42)
```


[1]: https://github.com/sanctuary-js/sanctuary-type-classes#filter
[2]: https://github.com/sanctuary-js/sanctuary-type-classes#filterM

TheLudd:
> I miss good way to wrap nullable values in the Maybe Monad.

I understand this as wrapping a "primitive" value, i.e. not a `Maybe`, in a maybe. I don't see that in your example.

davidchambers:
I see now that I wasn't clear, @TheLudd. What I'm suggesting is unconditionally applying `Just` to a value of type `a` to produce a value of type `Maybe a`, then filtering to replace an unwanted value with `Nothing`.

```javascript
Z.filter(S.even, Just(x));  // => Just(x) or Nothing, depending on x
```

One could define `nothingIf` in terms of these simpler functions:

```javascript
//    nothingIf :: (a -> Boolean) -> a -> Maybe a
[...]

PR #133, opened 2016-11-07
Added toMaybe static constructor.

martinlechner1:
As discussed in Issue #132, I added a proposal for a ofNullable static constructor.

### DIFF
--- a/docs/Maybe.md
+++ b/docs/Maybe.md
@@ -19,7 +19,7 @@ const Nothing = M.Nothing;
 
 const safeDiv = R.curry((n, d) => d === 0 ? Nothing() : Just(n / d));
 const lookup = R.curry((k, obj) => k in obj ? Just(obj[k]) : Nothing());
-const safeOperationWrapper = Maybe.ofNullable(unsafeOperation());
+const safeOperationWrapper = Maybe.toMaybe(unsafeOperation());
 ```
 
 ## Interaction
@@ -92,9 +92,9 @@ default value if a `Nothing` is received.
 Produces a pure `Maybe` instance of a given value. Effectively the `Just`
 constructor.
 
-#### `Maybe.ofNullable`
+#### `Maybe.toMaybe`
 ```hs
-:: a -> Maybe a
+:: a? -> Maybe a
 ```
 Produces a `Maybe` instance of a given value. Returns `Nothing` for nil values
 (undefined and null) and `Just(a)` for non-nil values.
--- a/docs/Maybe.md
+++ b/docs/Maybe.md
@@ -19,6 +19,7 @@ const Nothing = M.Nothing;
 
 const safeDiv = R.curry((n, d) => d === 0 ? Nothing() : Just(n / d));
 const lookup = R.curry((k, obj) => k in obj ? Just(obj[k]) : Nothing());
+const safeOperationWrapper = Maybe.ofNullable(unsafeOperation());
 ```
 
 ## Interaction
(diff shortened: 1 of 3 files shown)


davidchambers:
This function does not actually take a value of type [`Nullable a`][1] as an argument, since it also treats `undefined` specially. It's misleading to use _nullable_ in this function's name. I suggest `toMaybe`, to match [`S.toMaybe`][2].

I've been using `a?` to represent the type whose members are `null` and `undefined` in addition to all the members of `a`.


[1]: https://github.com/sanctuary-js/sanctuary-def#nullable
[2]: https://sanctuary.js.org/#toMaybe

martinlechner1:
Thanks for the input. I will change the code accordingly.

davidchambers:
The changes look good. Could you update the pull request's title accordingly?

TheLudd:
This is essentially an alias of the constructor. What about ramdas no alias policy?

CrossEye:
> This is essentially an alias of the constructor.

I don't think so.  You can't use a constructor (directly) as a pure function.  This is not a simple alias.

martinlechner1:
The primary motivation was to not use the constructor directly to have a cleaner general code style while using the lib.
See: http://eslint.org/docs/rules/new-cap
Having Maybe.toMaybe(value) is in my opinion superior to Maybe(value), since you express your intent more cleanly.

TheLudd:
I disagree with the claim that `toMaybe` is more clear. That name does not reveal anything except that it returns a Maybe. 

Regarding the rule, I have been forced to deactivate it on some projects since there are functions that are not constructors but that do start with a capital letter such as the `I`, `K` combinators.

CrossEye:
One more reason I don't like the single-letter combinator names!  :smile:

TheLudd:
> One more reason I don't like the single-letter combinator names!

Haha to each his own but I prefer eslint to be a helper and not rule my world =)

CrossEye:
No, I turn off linting rules all the time.  I've enforced similar rules in JS codebases long before I had linters available; it has seemed a good convention for JS.  

My main objection to the single-letter names is how nondescriptive they are.   I have the same problem using bird names.

martinlechner1:
What about the merge request? There is another not mentioned benefit. The classic constructor does not handle the undefined case (which I would like to have handled too).

TheLudd:
> The classic constructor does not handle the undefined case

yes it does. `x == null` equals true for `undefined` and `null`

src/Maybe.js
```diff
@@ -37,6 +39,8 @@ Maybe.of = Maybe.Just;
 
 Maybe.prototype.of = Maybe.Just;
 
+Maybe.toMaybe = ifElse(isNil, Maybe.Nothing, Maybe.Just);
```
davidchambers: Let's simplify the implementation:

```javascript
Maybe.toMaybe = Maybe;
```

test/maybe.test.js
```diff
@@ -222,4 +222,20 @@ describe('Maybe usage', function() {
...
+    it('returns Nothing for undefined', function() {
+      assert.equal(Maybe.toMaybe(undefined).isNothing, true);
+    });
+    it('returns Just(x) for an arbitrary value x', function() {
+      var just1 = Maybe.toMaybe(1);
+      assert.equal(just1.isJust, true);
+      assert.equal(just1.getOrElse(42), 1);
+    });
+  });
+
```
davidchambers: If we simplify the implementation as I suggested in my previous comment, it would be sufficient to assert that `Maybe.toMaybe === Maybe`. :)

docs/Maybe.md
```diff
@@ -91,6 +92,13 @@ default value if a `Nothing` is received.
 Produces a pure `Maybe` instance of a given value. Effectively the `Just`
 constructor.
 
+#### `Maybe.toMaybe`
+```hs
+:: a? -> Maybe a
+```
+Produces a `Maybe` instance of a given value. Returns `Nothing` for nil values
+(undefined and null) and `Just(a)` for non-nil values.
```
davidchambers: `Just(a)` is a bit confusing, as `a` is a type variable. How about this?

_Alias for `Maybe`. Produces a `Maybe` instance of a given value. Returns `Nothing` for `null`/`undefined`; a `Just` for any other value._

scott-christopher:
Closing as implemented via #159.

PR #141, opened 2016-12-25
Remove unnecessary currying

dmitriz:
There seems to be no need to curry in the examples, if I understand correctly.


davidchambers:
Currying the functions at point of definition seems more elegant to me. Let's keep it as is. :)

PR #142, opened 2017-01-03
Add State to Table on README

jonboiser:
Since it was missing. I based the checkmarks on the docs, let me know if I got something wrong.


scott-christopher:
Thanks @jonboiser 

🍄

PR #145, opened 2017-01-14
Create a join method

idcmardelplata:
I think it would be good for the types to have a join method. Here is a simple implementation.

Obviously used like this:

```js
Maybe.prototype.join = getJoinForTypes(Maybe);
IO.prototype.join = getJoinForTypes(IO)
etc...
```

### DIFF
--- a/src/internal/util.js
+++ b/src/internal/util.js
@@ -62,6 +62,15 @@ module.exports = {
         return Type.of(f(a));
       });
     };
+  },
+  
+  getJoinForTypes : function(m) {
+  return function() {
+    return this.value instanceof m
+      ? this.value.join()
+      : m.of(this.value);
+    }
   }
+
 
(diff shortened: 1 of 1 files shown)


davidchambers:
Are you imagining something like `join :: Chain m => m (m a) -> m a`, @idcmardelplata?

idcmardelplata:
As it is, I do not understand why the types do not have method join.

davidchambers:
[`join`][1] can be derived from `chain`. In my view we're better off defining `join` once and for all than having dozens of authors add `join` methods to their algebraic data types.


[1]: https://github.com/sanctuary-js/sanctuary-type-classes/blob/v2.0.1/index.js#L1250-L1268

idcmardelplata:
As such, I think the same as you

Issue #153, opened 2017-03-04
Either.fold


OliverJAsh:
Are there any plans to add `fold` methods to these data structures, [like in Folktale](http://docs.folktalejs.org/en/latest/api/data/either/Either.html)?

ferborva:
Dispite there being a static function available which works the same way as `fold`:

Either.either(leftTransformation, righTransformation, EitherInstance)

it might be interesting to have the a method attached to the prototype which does the same thing.

OliverJAsh:
That's good to know, thanks!

Are we suggesting it should implement Foldable?

PR #155, opened 2017-03-06
Either instance 'either' method

ferborva:
In reference to the comment made by @OliverJAsh on [#153](https://github.com/ramda/ramda-fantasy/issues/153) I thought it might be a simple addition and interesting one to have on the either prototype.

### DIFF
--- a/src/Either.js
+++ b/src/Either.js
@@ -127,4 +127,9 @@ Either.Left = function(value) {
 };
 
 
+// either
+Either.prototype.either = function instanceEither(leftFn, rightFn) {
+  return this.isLeft ? leftFn(this.value) : rightFn(this.value);
+};
+
 module.exports = Either;
--- a/docs/Either.md
+++ b/docs/Either.md
@@ -140,3 +140,10 @@ are of the same constructor and both contain equal values.
 :: Either a b ~> () -> String
 ```
 Returns a string representation of the `Either` instance.
+
+#### `either.either`
+```hs
+:: Either a b ~> (a -> c) -> (b -> c) -> c
+```
+Used to extract the value out of the `Either` by providing a function to handle
+the types of values contained in both `Left` and `Right`.
--- a/test/either.test.js
+++ b/test/either.test.js
@@ -199,4 +199,20 @@ describe('Either', function() {
 
   });
 
+
+
+  describe('#either', function() {
+    it('returns the value of a Left after applying the first function arg', function() {
+      jsv.assert(jsv.forall(leftArb(jsv.nat), fnNatArb, fnNatArb, function(e, f, g) {
+        return e.either(f, g) === f(e.value);
+      }));
+    });
+
+    it('returns the value of a Right after applying the second function arg', function() {
+      jsv.assert(jsv.forall(rightArb(jsv.nat), fnNatArb, fnNatArb, function(e, f, g) {
+        return e.either(f, g) === g(e.value);
+      }));
+    });
+  });
+
 });


src/Either.js
```diff
@@ -127,4 +127,12 @@ Either.Left = function(value) {
 };
 
 
+// Fold methods
+_Right.prototype.fold = _Left.prototype.fold = function eitherFold(leftFn, rightFn) {
```
buzzdecafe: why not just add it to `Either.prototype` and let `Left` and `Right` inherit?

ferborva: Indeed

src/Either.js
```diff
@@ -127,4 +127,12 @@ Either.Left = function(value) {
 };
 
 
+// Fold
+Either.prototype.fold = function eitherFold(leftFn, rightFn) {
+  if (this instanceof _Left) {
+    return leftFn(this.value);
+  }
+  return rightFn(this.value);
+};
```
buzzdecafe: rather than relying on `instanceof`, let's use the `isLeft` property (which sanctuary also uses), and write this as a ternary expression, e.g.

`return (this.isLeft) ? leftFn(this.value) : rightFn(this.value);`

davidchambers: Or even this:

```javascript
return (this.isLeft ? leftFn : rightFn)(this.value);
```

bijoythomas: Why not re-use the static `Either.either` function and further abstract the check for Left or Right?

```
Either.prototype.fold = function(leftFn, rightFn) {
  return Either.either(leftFn, rightFn)(this)
}
```

bijoythomas:
Also, based on the spec here: https://github.com/fantasyland/fantasy-land#foldable, shouldn't the method be called `reduce` (like the Maybe type) instead of fold?

ferborva:
Yep. To stay in line with fantasyland spec it should be called 'reduce'. On the other hand, using Either.either ... don't know what to say to you on that one. Personally I don't see too much use in it to start with as it seems for reasonable to unfold the either from the instance itself rather than using a static method.

davidchambers:
Start with the general type of [`reduce`][1]:

```haskell
reduce :: Foldable f => f a ~> ((b, a) -> b, b) -> b
```

Replace `Foldable f => f` with `Either x`:

```haskell
reduce :: Either x a ~> ((b, a) -> b, b) -> b
```

So, the method must take a `(b, a) -> b` function and a value of type `b`, and return a `b`. Given `Left('XXX')` we do not have a value of type `b` so our only option is to return the `b` provided to us. Given `Right(42)` we have two options: return the given `b`, or apply the provided function to the given `b` and `42` and return the resulting `b`. The latter is far more useful.

The implementation of [`Either#fantasy-land/reduce`][2] looks something like this:

> ```javascript
> Either.prototype['fantasy-land/reduce'] = function(f, x) {
>   return this.isRight ? f(x, this.value) : x;
> };
> ```

Note that `reduce` cannot operate on the value inside a Left. The function being defined in this pull request is something altogether different.


[1]: https://github.com/fantasyland/fantasy-land#reduce-method
[2]: https://github.com/sanctuary-js/sanctuary/blob/v0.12.2/index.js#L2210-L2228

ferborva:
Meaning that adding the .either static method (as .fold or ... whatever) to the Either instance would not make it Foldable ... Can't see this signature mapped on any of the fantasyland specs: Either a b ~> (a -> c) -> (b -> c) -> c.

bijoythomas:
Thanks for the explanation @davidchambers. But then, why is there an Either.either that allows running a leftFn on the value of a Left instance? 

I can see how for a Maybe.Nothing (which is the absence of a value), having reduce return the default makes sense. But a Left instance does not indicate the absence of a value.

davidchambers:
[`either`][1] is a completely valid function, @bijoythomas. It's just a different function from `reduce`. There's no way to make the types line up.


[1]: http://hackage.haskell.org/package/base-4.9.1.0/docs/Prelude.html#v:either

davidchambers:
Would you mind squashing these commits, @ferborva?

ferborva:
@davidchambers No problem :)

davidchambers:
Thank you, @ferborva. The commit message still refers to `fold`. Could you update it?

README.md
```diff
@@ -12,7 +12,7 @@ not undergone thorough testing/use yet.
 
 | Name            | [Setoid][3]  | [Semigroup][4] | [Functor][5] | [Applicative][6] | [Monad][7] | [Foldable][8] | [ChainRec][16] |
 | --------------- | :----------: | :------------: | :----------: | :--------------: | :--------: | :-----------: | :------------: |
-| [Either][9]     |    **✔︎**     |                |     **✔︎**    |      **✔︎**       |   **✔︎**    |               |     **✔︎**      |
+| [Either][9]     |    **✔︎**     |                |     **✔︎**    |      **✔︎**       |   **✔︎**    |              |     **✔︎**      |
```
davidchambers: Please revert this change.

ferborva:
@davidchambers There we go

PR #158, opened 2017-04-08
docs/Reader: fix possibly incorrect type sigs

raine:
I see `ReaderT` being returned elsewhere so I assume these should return `ReaderT` as well.


buzzdecafe:
🐄

PR #167, opened 2017-11-28
Fixing vulnerability with old version of uglify

CrossEye:


davidchambers:
Will you publish a new release?

Bradcomp:
:bowtie:

CrossEye:
I think we'll have to, much as I don't really like to do anything with this.

I've done little with R-F.  Can anyone tell me if we've got the release process documented?

davidchambers:
> Can anyone tell me if we've got the release process documented?

`npm run release-patch` should do the trick.

CrossEye:
Yes, I can build and tag, but I guess I was never added to the list of npm authors.  Could someone who has access either add me or do a `publish`? 

@ramda/core @ramda/ramda-fantasy

davidchambers:
I don't think a new release is necessary actually. As I understand it the UglifyJS vulnerability affects the `uglifyjs` program itself, not its output. Users of ramda-fantasy are unaffected.

CrossEye:
Well, the npm package includes everything, `src`, `test`, etc.  Anyone could build from there.

So it seems safer to publish the updated version.

